PR #3, opened 2020-05-19
add Rabbitmq/amqp source

sbcd90:
This source integrates Knative eventing with Rabbitmq which supports AMQP protocol. The RabbitmqSource has the ability to consume messages that are emitted from a Rabbitmq Producer using AMQP protocol. After consuming a message, the RabbitmqSource will post the retrieved value to an assigned sink such as a channel or service.

This RabbitmqSource has the following features:

Supports Basic Authentication
Supports pub/sub, routing keys & topic based messaging


source/pkg/reconciler/rabbitmqsource.go
```diff
@@ -0,0 +1,241 @@
...
+	duckv1 "knative.dev/pkg/apis/duck/v1"
+	"knative.dev/pkg/logging"
+	pkgLogging "knative.dev/pkg/logging"
+	"knative.dev/pkg/metrics"
+	pkgreconciler "knative.dev/pkg/reconciler"
+	"knative.dev/pkg/resolver"
+)
+
+const (
+	raImageEnvVar              		= "RABBITMQ_RA_IMAGE"
```
vaikas, line 48: nit, indent is off here and few other places. Maybe run go fmt?

source/pkg/reconciler/rabbitmqsource.go
```diff
@@ -0,0 +1,241 @@
...
+	loggingContext		context.Context
+	loggingConfig 		*pkgLogging.Config
+	metricsConfig 		*metrics.ExporterOptions
+
+	sinkResolver		*resolver.URIResolver
+}
+
+var _ reconcilerrabbitmqsource.Interface  = (*Reconciler)(nil)
+
+func (r *Reconciler) ReconcileKind(ctx context.Context, src *v1alpha1.RabbitmqSource) pkgreconciler.Event {
```
vaikas, line 93: Don't you also need to have FinalizeKind for any cleanup? Just jotting this down so I don't forget :)

source/pkg/reconciler/resources/receive_adapter.go
```diff
@@ -0,0 +1,181 @@
...
+	}
+
+	return &v1.Deployment{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:         utils.GenerateFixedName(args.Source, fmt.Sprintf("rabbitmqsource-%s", args.Source.Name)),
+			Namespace:    args.Source.Namespace,
+			GenerateName: fmt.Sprintf("%s-", args.Source.Name),
+			Labels:       args.Labels,
+			OwnerReferences: []metav1.OwnerReference{
+				*kmeta.NewControllerRef(args.Source),
```
vaikas, line 152: if this is the only thing you create then you don't need finalize, ack 👍

source/pkg/reconciler/resources/receive_adapter.go
```diff
@@ -0,0 +1,181 @@
...
+			Value: args.Source.Spec.Topic,
+		},
+		{
+			Name: "RABBITMQ_USER",
+			ValueFrom: &corev1.EnvVarSource{
+				SecretKeyRef: args.Source.Spec.User.SecretKeyRef,
+			},
+		},
+		{
+			Name: "RABBITMQ_PASSWORD",
```
vaikas, line 58: it would be better if creds came from a mounted volume, see the comment earlier about bindings.

source/pkg/reconciler/resources/receive_adapter.go
```diff
@@ -0,0 +1,181 @@
...
+		},
+		Spec: v1.DeploymentSpec{
+			Selector: &metav1.LabelSelector{
+				MatchLabels: args.Labels,
+			},
+			Replicas: &replicas,
+			Template: corev1.PodTemplateSpec{
+				ObjectMeta: metav1.ObjectMeta{
+					Annotations: map[string]string{
+						"sidecar.istio.io/inject": "true",
```
vaikas, line 163: Do we really want to inject istio always?

go.mod
```diff
@@ -0,0 +1,30 @@
+module knative.dev/eventing-rabbitmq
+
+go 1.13
+
+require (
+	github.com/cloudevents/sdk-go v1.2.0
+	github.com/cloudevents/sdk-go/v2 v2.0.0-RC4
+	github.com/google/ko v0.5.1 // indirect
+	github.com/robfig/cron v1.2.0 // indirect
+	github.com/sbcd90/wabbit v0.0.0-20190419210920-43bc2261e0e0
```
markfisher, line 10: will the changes on your fork be merged into the main project?

hack/OWNERS
```diff
@@ -0,0 +1,10 @@
+# The OWNERS file is used by prow to automatically merge approved PRs.
+
+approvers:
+- productivity-approvers
```
markfisher, line 4: should this be eventing approvers, at least initially?

hack/release.sh
```diff
@@ -0,0 +1,52 @@
+#!/usr/bin/env bash
+
+# Copyright 2018 The Knative Authors
```
markfisher, line 3: 2020

hack/release.sh
```diff
@@ -0,0 +1,52 @@
...
+export GO111MODULE=on
+
+# Yaml files to generate, and the source config dir for them.
+declare -A COMPONENTS
+COMPONENTS=(
+  ["rabbitmq.yaml"]="source/config"
+)
+readonly COMPONENTS
+
+function build_release() {
```
markfisher, line 31: this is still referring to `contrib.eventing`

hack/update-codegen.sh
```diff
@@ -0,0 +1,59 @@
+#!/usr/bin/env bash
+
+# Copyright 2018 The Knative Authors
```
markfisher, line 3: 2020

hack/update-deps.sh
```diff
@@ -0,0 +1,66 @@
+#!/usr/bin/env bash
+
+# Copyright 2018 The Knative Authors
```
markfisher, line 3: 2020

source/PROJECT
```diff
@@ -0,0 +1,3 @@
+version: "1"
+domain: eventing.knative.dev
+repo: knative.dev/eventing-contrib
```
markfisher, line 3: should be `eventing-rabbitmq`

source/config/201-clusterrole.yaml
```diff
@@ -0,0 +1,82 @@
...
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  name: eventing-sources-rabbitmq-controller
+  labels:
+    contrib.eventing.knative.dev/release: devel
+rules:
+
+  - apiGroups:
+      - sources.knative.dev
```
markfisher, line 24: should projects in the sandbox continue to use the common api groups or define their own?
cc @vaikas @evankanderson

source/pkg/adapter/adapter.go
```diff
@@ -0,0 +1,351 @@
...
+		httpMessageSender: httpMessageSender,
+		reporter:		   reporter,
+		logger: 		   logger,
+		context: 		   ctx,
+	}
+}
+
+func (a *Adapter) CreateConn(User string, Password string, logger *zap.Logger) (*amqp.Conn, error) {
+	if User != "" && Password != "" {
+		a.config.Brokers = fmt.Sprintf("amqp://%s:%s@%s", a.config.User, a.config.Password, a.config.Brokers)
```
markfisher, line 108: This will be logged in plaintext.

source/pkg/adapter/adapter.go
```diff
@@ -0,0 +1,351 @@
...
+				},)
+			if err != nil {
+				logger.Error(err.Error())
+				return nil, err
+			}
+		}
+	} else {
+		err = (*ch).QueueBind(
+			queue.Name(),
+			"",
```
markfisher, line 212: should be passing the `QueueConfig.RoutingKey` here, right?

source/pkg/adapter/adapter.go
```diff
@@ -0,0 +1,351 @@
...
+	req, err := a.httpMessageSender.NewCloudEventRequest(a.context)
+	if err != nil {
+		return err
+	}
+
+	event := cloudevents.NewEvent()
+
+	event.SetID(msg.MessageId())
+	event.SetTime(msg.Timestamp())
+	event.SetType(sourcesv1alpha1.RabbitmqEventType)
```
markfisher, line 292: Maybe not in this initial PR, but there would ideally be a way for a user header to specify the event type (since rabbitmq is typically just going to be the transport, not the "type"). This might lead to a discussion of a CloudEvent binding for amqp 0.91.

source/pkg/apis/sources/v1alpha1/rabbitmq_types.go
```diff
@@ -0,0 +1,169 @@
...
+	NoWait           bool   `json:"nowait,omitempty"`
+}
+
+type RabbitmqSourceSpec struct {
+	// Brokers are the Rabbitmq servers the consumer will connect to.
+	// +required
+	Brokers string `json:"brokers"`
+	// Topic topic to consume messages from
+	// +required
+	Topic   string `json:"topic,omitempty"`
```
markfisher, line 113: why is this here? aren't message consumed from the Queue that is bound to the Exchange?

source/pkg/reconciler/resources/receive_adapter.go
markfisher: well, there will also typically be resources created on the RabbitMQ cluster as well (Exchange and Queue, unless they were pre-existing) and the Binding from the Exchange to the Queue. Those need to be cleaned up on deletion as well, and the Binding might need to be replaced if there is an update to source properties that affect it

source/config/201-clusterrole.yaml
evankanderson: Good question. There's been some [discussion about that in the TOC and Steering](https://docs.google.com/document/d/1e0PpX4uUaaolF7k5p05ZK_ScjpS5VRiK5y8cCg4oH7Q/edit#heading=h.ya000o6lypo9); the current proposal is to use a separate domain for the CRD APIs, but I don't think we have such a subdomain yet.

In this case, I'd suggest begging forgiveness and alignment with the other sources.

source/pkg/reconciler/resources/receive_adapter.go
vaikas: Yeah, that's why I tagged it first above since I was expecting there to be some sort of cleanup necessary, but I must have missed (as I said, just the first pass) on what other cleanup needs to happen. Maybe you could add a comment in the PR at the location where the creation of resources happens that need to get cleaned up, so it won't get lost?

source/pkg/adapter/adapter.go
```diff
@@ -0,0 +1,351 @@
...
+	}
+
+	return a.PollForMessages(&ch, queue, stopCh)
+}
+
+func (a *Adapter) StartAmqpClient(ch *wabbit.Channel) (*wabbit.Queue, error) {
+	logger := a.logger
+	exchangeConfig := fillDefaultValuesForExchangeConfig(&a.config.ExchangeConfig, a.config.Topic)
+
+	err := (*ch).ExchangeDeclare(
```
markfisher, line 166: Perhaps the lifecycle of these resources (Exchange, Queue and the Bindings below) should be managed by the reconciler? That would allow for updates/deletes to occur there as well. This code could then be changed to use "passive" versions of the declare operations.

As far as I can tell, in the current implementation, even if a Source resource is deleted, these resources will remain (which means the Queue continues to receive messages via the Binding).

vaikas: Cool, thanks 👍

vaikas:
@sbcd90 Do you have a rough idea when you might be able to take a spin at the suggestions, or what do you think about them?

sbcd90:
@vaikas , thanks a lot for pinging me. i should be able to address most of them by this weekend i.e. May 31st.

sbcd90:
Hi @vaikas , @markfisher , Thanks a lot for your review comments. I have addressed the smaller ones while the big changes are still not resolved.
Is it possible that this PR is merged & new issues are created based on the remaining comments & I keep on creating individual PRs for them. It will be easier for me.
Please let me know.

vaikas:
/lgtm
/approve

Sure thing, let's get this in and then work through things in follow up PRs 👍

PR #17, opened 2020-09-01
gofmt

vaikas:
Ran:
```
for i in `find . -path './vendor' -prune -o -path './third_party' -prune -o -type f -name '*.go' -print`; do gofmt -s -w $i; done
```


vaikas:
FYI, you can see them here?
https://github.com/knative-sandbox/eventing-rabbitmq/actions

n3wscott:
/lgtm
/unhold

We will sort the verify out in a followup

PR #44, opened 2020-09-10
only reconcile my own triggers

vaikas:
Only reconcile my triggers. While we only enqueue our triggers on broker changes, during global resync they would also get queued.
While we're migrating to new Keda, disable it until things settle down there. Will be apparently fixed in upstream in a matter of days...


n3wscott (commented):
/lgtm
/approve

PR #49, opened 2020-09-10
move broker config files to better location

vaikas:
move files from config/config => config/broker & update readme (which was wrong to begin with).
Also keda is moving to a new apigroup, so add rights to it as well (keda.sh).


n3wscott (commented):
/lgtm
/approve

PR #50, opened 2020-09-10
have one replica for dispatcher until we get keda back up

vaikas:
Spin the dispatcher to one replica until we sort keda out.


n3wscott (commented):
/lgtm
/approve

/hold
I want to see the e2e run

PR #62, opened 2020-09-17
consistent queue names, v1.Trigger

vaikas:
Fallout from using Keda because we had to change the queue names, so make sure they are consistent.
Also use v1.Trigger instead of v1beta1.


n3wscott (commented):
/lgtm
/approve

PR #76, opened 2020-09-22
Update README.md

Gsantomaggio:
Create the required namespace `knative-eventing`
fix typo in `ko`

### DIFF
--- a/broker/README.md
+++ b/broker/README.md
@@ -42,7 +42,8 @@ install KEDA as documented [here](https://keda.sh/docs/latest/deploy/).
 install the broker-controller from this repository:
 
 ```
-ko apply -f config/broker/
+kubectl create namespace knative-eventing
+kubectl apply -f config/broker/
 ```
 
 ## Demo


broker/README.md
```diff
@@ -42,7 +42,8 @@ install KEDA as documented [here](https://keda.sh/docs/latest/deploy/).
 install the broker-controller from this repository:
 
 ```
-ko apply -f config/broker/
+kubectl create namespace knative-eventing
```
vaikas, line 45: This should have been created for you as per L7 (installing eventing creates the knative-eventing namespace).

broker/README.md
```diff
@@ -42,7 +42,8 @@ install KEDA as documented [here](https://keda.sh/docs/latest/deploy/).
...
+kubectl apply -f config/broker/
```
vaikas, line 46: This is actually not going to work with just plain kubectl. You have to use ko because the images are using ko:// so unless you use ko they won't get built and the controller won't start, well, nothing will work :)

vaikas:
We should have a release shortly, and then we can use kubectl apply -f http://... without requiring ko.

vaikas:
Now having said that, ugh, the readme is out of date (been a busy few days :) ) so I had on my TODO to fix it. Sorry about that.

Gsantomaggio:
@vaikas Thank you for the clarification. I am going to close it.

I think that also this https://github.com/knative-sandbox/eventing-rabbitmq/issues/77 is because the documentation is out of date.

PR #107, opened 2020-10-06
use the passed in duration, clean readme

vaikas:
Turns out the cloudevents sdk does NOT retry on 500, so changed the example to 503 to show the retries. Also, use the passed in duration instead of ignoring it.

### DIFF
--- a/samples/broker/dlq/README.md
+++ b/samples/broker/dlq/README.md
@@ -225,9 +225,9 @@ vaikas-a01:wabbit vaikas$ kubectl -n dlq-demo -l='serving.knative.dev/service=fa
 ```
 
 You see there are both 200 / 503 events there. And more importantly, you can see
-that 200 is only sent once to the failer since it's processed
-correctly. However, the 503 is sent a total of 6 times because we have specified
-the retry of 5 (original, plus 5 retries for a total of 6 log entries).
+that 200 is only sent once to the failer since it's processed correctly.
+However, the 503 is sent a total of 6 times because we have specified the retry
+of 5 (original, plus 5 retries for a total of 6 log entries).
 
 However the event-display (the Dead Letter Sink) only sees the failed events
 with the response code set to 503.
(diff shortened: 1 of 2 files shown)


samples/broker/dlq/README.md
```diff
@@ -206,41 +206,45 @@ EOF
...
+2020/10/06 10:35:01 using response code: 503
+2020/10/06 10:35:01 using response code: 503
+2020/10/06 10:35:03 using response code: 503
 ```
 
-You see there are both 200 / 500 events there.
+You see there are both 200 / 503 events there. And more importanly, you can see
+that 200 is only sent once to the failer since it's processed
+correctly. However, the 503 is sent a total of 6 times because we have specified
+the retry of 5 (original, plus 5 retries for a total of 6 log entries).
```
mattmoor, line 230: Format markdown:
```suggestion
that 200 is only sent once to the failer since it's processed correctly.
However, the 503 is sent a total of 6 times because we have specified the retry
of 5 (original, plus 5 retries for a total of 6 log entries).
```

mattmoor (commented):
Produced via:
  `prettier --write --prose-wrap=always $(find -name '*.md' | grep -v vendor | grep -v .github | grep -v docs/cmd/)`

samples/broker/dlq/README.md
```diff
@@ -206,41 +206,45 @@ EOF
...
+2020/10/06 10:35:01 using response code: 503
+2020/10/06 10:35:01 using response code: 503
+2020/10/06 10:35:03 using response code: 503
 ```
 
-You see there are both 200 / 500 events there.
+You see there are both 200 / 503 events there. And more importantly, you can see
+that 200 is only sent once to the failer since it's processed
+correctly. However, the 503 is sent a total of 6 times because we have specified
+the retry of 5 (original, plus 5 retries for a total of 6 log entries).
```
mattmoor, line 230: Format markdown:
```suggestion
that 200 is only sent once to the failer since it's processed correctly.
However, the 503 is sent a total of 6 times because we have specified the retry
of 5 (original, plus 5 retries for a total of 6 log entries).
```

mattmoor (commented):
Produced via:
  `prettier --write --prose-wrap=always $(find -name '*.md' | grep -v vendor | grep -v .github | grep -v docs/cmd/)`

n3wscott (commented):
/lgtm
/approve

PR #113, opened 2020-10-08
Use a custom retriable function

vaikas:
By default cloudevents client has a very limited set of response codes it retries (for example, not 500), so use a custom method now that this landed:
https://github.com/cloudevents/sdk-go/pull/596

And then this :)
https://github.com/cloudevents/sdk-go/pull/597


pkg/dispatcher/dispatcher.go
```diff
@@ -130,6 +128,10 @@ func (d *Dispatcher) ConsumeFromQueue(ctx context.Context, channel wabbit.Channe
 	return nil
 }
 
+func isRetriableFunc(sc int) bool {
+	return sc < 200 || sc >= 300
```
slinkydeveloper, line 132: Retry all the things!

slinkydeveloper (approved):
/lgtm

PR #147, opened 2020-10-30
Recordevents move

vaikas:
https://github.com/knative/eventing/pull/4406
removes these old locations, so fix the locations to new place.


vaikas:
/hold
Actually I think we should get rid of this and just use a container that gets published from eventing or maybe further along from reconciler-testing.

n3wscott (commented):
/lgtm
/approve

vaikas:
Closing in favor of #189

PR #159, opened 2020-11-10
exercise more eventing-versions

### LINKED ISSUE #155: Add e2e tests with different versions of eventing.
We currently hard code the version of eventing we install as discovered in #153. We should probably have a matrix tests so we test for example, .17, .18 and .19 once it comes out.

vaikas:
Fixes #155


n3wscott (commented):
/lgtm
/approve

PR #177, opened 2020-12-09
no underscore in name

vaikas:
# Changes
Change the controller name, which was causing failures when trying to get a leader election lock.

-
-
-

/kind <kind>

Fixes #

**Release Note**

```release-note

```

**Docs**

```docs

```


mattmoor (approved):
/lgtm

PR #180, opened 2020-12-21
add e2e tests do not always json encode input

vaikas:
# Add e2e tests for source using Kind

- :gift: Add e2e tests for source using Kind
- :bug: Do not always try to json encode the input
- :broom: Update outdated documentation, about install, etc.

/kind <kind>

Fixes #

**Release Note**

```release-note

```

**Docs**

```docs

```


vaikas:
/close
#183 is new version from my repo instead of (mistakenly set to this one).

PR #229, opened 2021-02-16
Fail releasability if NO-GO

mattmoor:
If this WAI, then I'll update .github with the same change.

### DIFF
--- a/.github/workflows/knative-releasability.yaml
+++ b/.github/workflows/knative-releasability.yaml
@@ -157,3 +157,11 @@ jobs:
             ${{ env.CHECK_MESSAGE }}
             ${{ env.VERIFY_MESSAGE }}
             For detailed logs: https://github.com/${{ github.repository }}/actions/runs/${{ github.run_id }}
+
+      - name: Fail if NO-GO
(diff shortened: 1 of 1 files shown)


n3wscott (commented):
/lgtm
/approve

.github/workflows/knative-releasability.yaml
```diff
@@ -157,3 +157,11 @@ jobs:
             ${{ env.CHECK_MESSAGE }}
             ${{ env.VERIFY_MESSAGE }}
             For detailed logs: https://github.com/${{ github.repository }}/actions/runs/${{ github.run_id }}
+
+      - name: Fail if NO-GO
+        if: steps.exists.outputs.release-branch == 'false' && env.current == 'false'
```
vaikas, line 162: @mattmoor this is same condition as on L142. Just curious if setting a different env variable in that step and then checking it and if it's failed, exit 1 there after sending the slack message would have the same effect? Only thinking in keeping them in sync might be easier that way.

mattmoor: It should have the same effect, though I'm not sure it's any simpler.  This is now propagated everywhere, so if we wanna change it we should do it upstream in `knative-sandbox/.github`

PR #258, opened 2021-03-17
test with less, seems we are getting throttled

vaikas:
# Changes

- :broom: Seems like we're getting throttled, drop some of the matrix components to reduce the load.
* Do not test eventing .18.X anymore
* Do not test with rabbitmq cluster operator {v1.[0,1,2],0}

/kind <kind>

Fixes #

**Release Note**

```release-note

```

**Docs**

```docs

```


n3wscott (approved):
/lgtm

rofl

PR #271, opened 2021-04-08
Use CRDs instead of client libraries with RabbitMQ cluster operator

vaikas:
# Changes

- :gift: Start using the messaging-topology-operator CRDs for creating RabbitMQ resources if using the RabbitMQ cluster operator.
-
-

/kind <kind>

Fixes #

**Release Note**

```release-note
**NOTE**  If using RabbitMQ Cluster Operator, requires > 1.6. If you are using secret based (and hence libraries like rabbithole), no changes there then.
:bug: If a Broker / Trigger is deleted but there's no secret to talk to the rabbit cluster, we now remove the resources and log that rabbitmq resources _might_ have been leaked. This happens only when not using the Rabbit Cluster Operator. This left undeletable resources, so this change seems more user friendly.
```

**Docs**

```docs

```


n3wscott (approved):
/lgtm
/approve

PR #280, opened 2021-05-04
rework the ingress of broker for rabbit

n3wscott:
Found several issues looking at the implementation of the ingress, notably it was not rejecting non-POST methods, and a few other various return code control things we are suppose to do for passing conformance on Broker specs.

This pulls in the http listener that eventing MTChannelBasedBroker classed brokers use and then pushes the event out onto the rabbit channel like it did before.

I also increased the polling timeout, I found the default 2 minutes is on the edge of flakey for RabbitMQ on azure. The PVC seems to be pretty slow on azure, and also could be a source of issue for the flakes we are seeing. Bumped to 5 minutes.

### DIFF
--- a/cmd/ingress/main.go
+++ b/cmd/ingress/main.go
@@ -39,12 +39,12 @@ const (
 )
 
 type envConfig struct {
-	Port          int    `envconfig:"PORT" default:"8080"`
-	BrokerURL string `envconfig:"BROKER_URL" required:"true"`
+	Port         int    `envconfig:"PORT" default:"8080"`
+	BrokerURL    string `envconfig:"BROKER_URL" required:"true"`
 	ExchangeName string `envconfig:"EXCHANGE_NAME" required:"true"`
 
 	channel *amqp.Channel
-	logger *zap.SugaredLogger
+	logger  *zap.SugaredLogger
 }
 
 func main() {
--- a/cmd/ingress/main.go
+++ b/cmd/ingress/main.go
@@ -19,62 +19,132 @@ package main
 import (
 	"context"
 	"encoding/json"
+	"fmt"
 	"log"
-	"os"
+	"net/http"
 
 	cloudevents "github.com/cloudevents/sdk-go/v2"
+	"github.com/cloudevents/sdk-go/v2/binding"
+	cehttp "github.com/cloudevents/sdk-go/v2/protocol/http"
+	"github.com/kelseyhightower/envconfig"
 	"github.com/streadway/amqp"
+	"go.uber.org/zap"
+	"knative.dev/eventing/pkg/kncloudevents"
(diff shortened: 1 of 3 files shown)


cmd/ingress/main.go
```diff
@@ -19,62 +19,132 @@ package main
...
+	defaultMaxIdleConnections        = 1000
+	defaultMaxIdleConnectionsPerHost = 1000
 )
 
+type envConfig struct {
+	Port         int    `envconfig:"PORT" default:"8080"`
+	BrokerURL    string `envconfig:"BROKER_URL" required:"true"`
+	ExchangeName string `envconfig:"EXCHANGE_NAME" required:"true"`
+
+	channel *amqp.Channel
```
vaikas, line 46: Just curious on why to shove these into envConfig instead of keeping the separation here between incoming environmental variables.

cmd/ingress/main.go
```diff
@@ -19,62 +19,132 @@ package main
...
-			amqp.Publishing{
-				Headers:     headers,
-				ContentType: "application/json",
-				Body:        bytes,
-			})
-		if err != nil {
-			log.Fatal("failed to publish message")
-		}
-	}))
+func (env *envConfig) send(event *cloudevents.Event) (int, error) {
```
vaikas, line 124: I think I'd prefer the signature to have context, channel and exchangeName in it rather than pull them from the envConfig (see above comment).

vaikas: That way at least the context from the request could be wired through that you get on L97?

vaikas (commented):
Thanks for cleaning this up!

cmd/ingress/main.go
n3wscott: I was going to wire in the context, but I did not see a context being used here so I skipped it

cmd/ingress/main.go
n3wscott: because I needed an instance of a struct to wire the send function and ServeHTTP to and all this code is local to the main file here. The right way would be to move this into a pkg folder somewhere I think

cmd/ingress/main.go
n3wscott: I did it this way because I already have envconf overloaded with the `func (env *envConfig) ServeHTTP` function

cmd/ingress/main.go
vaikas: Right, that's what I was thinking as well (move to pkg) so if it's not in the envConfig now might make it easier.

vaikas (approved):
/lgtm
This is better, we can refactor later. Main thing with the context was that just wondering if that will make cleanups (clean exiting, handling timeouts, etc.) easier. But, followup.

Issue #281, opened 2021-05-04
Trigger dispatchers do not exit cleanly.


vaikas:
Looks like things are not exiting cleanly. Just jotting this down so we don't forget:
```
t1-cenpmhqu-dispatcher-7445cc894f-qctcw          0/1     CreateContainerConfigError   0          8m10s
t1-ldyylthv-dispatcher-5b977c46fd-ltwng          0/1     CreateContainerConfigError   0          11m
t1-oexecgju-dispatcher-54fcd9d68c-8r8bf          0/1     CrashLoopBackOff             0          72s
t1-ogstraar-dispatcher-54f49d5bbb-wh567          0/1     CrashLoopBackOff             0          4m44s
t1-stgmqson-dispatcher-8c97cc758-mjpvw           0/1     CreateContainerConfigError   0          6m37s
t1-xlsarhmz-dispatcher-7895d7b777-tbtvh          0/1     CreateContainerConfigError   0          10m
trigger-harowruw-dispatcher-6f6f559c7c-8d66b     0/1     CrashLoopBackOff             8          17m
trigger-kqjgoumk-dispatcher-6b8d5dc6d6-676hd     0/1     CrashLoopBackOff             8          17m
trigger-kuxgiowz-dispatcher-5594885888-djk6t     0/1     CrashLoopBackOff             8          17m
trigger-lhxdajgi-dispatcher-6cf4d66c6f-grj8m     0/1     CrashLoopBackOff             8          17m
trigger-ljxmbclt-dispatcher-8465fb4ccc-bmrr8     0/1     CrashLoopBackOff             8          17m
```

So they only seem to happen after something underneath them gets deleted, like so:
```
vaikas@villes-mbp eventing-rabbitmq % kubectl -n test-ejajrvpg logs t1-ogstraar-dispatcher-54f49d5bbb-wh567
{"level":"info","ts":1620152272.595133,"logger":"fallback","caller":"dispatcher/main.go:74","msg":"BackoffDelay is zero, enforcing default","default":0.05}
{"level":"info","ts":1620152272.5953605,"logger":"fallback","caller":"dispatcher/main.go:77","msg":"Setting BackoffDelay","backoffDelay":0.05}
{"level":"info","ts":1620152272.6099005,"logger":"fallback","caller":"dispatcher/dispatcher.go:85","msg":"rabbitmq receiver started, exit with CTRL+C"}
[...]

PR #282, opened 2021-05-04
do not warn if conn is already closed, normal

vaikas:
- :broom: Do not unnecessarily warn if closing the channel fails if it's already been closed (#281)
-
-

/kind <kind>

Fixes #

**Release Note**

```release-note

```

**Docs**

```docs

```


n3wscott (approved):
/lgtm
/approve

PR #288, opened 2021-05-05
Remove executable flag from non-executable files

benmoss:
# Changes

- :broom: Remove the executable flag from a bunch of random files

/kind cleanup

Not sure why all these files got marked executable, assume it was just a mistake.


n3wscott (approved):
/lgtm
/approve

PR #293, opened 2021-05-10
Crd ut 2

vaikas:
# Changes

- :gift: Add more Unit Tests.
- :broom: Remove some old TODOs, we do now set the condition properly for example.
- :broom: Hoist common code between CRD / Library impls into a method and use that.

/kind <kind>

Fixes #

**Release Note**

```release-note

```

**Docs**

```docs

```


n3wscott (approved):
/lgtm
/approve

Issue #304, opened 2021-05-17
Add validating webhook targeting our Broker class.


vaikas:
We should add a webhook that checks incoming Brokers and makes sure that if they have our BrokerClass, that we validate it, so we can move code like this out of the backend and instead fail fast.

https://github.com/knative-sandbox/eventing-rabbitmq/blob/main/pkg/reconciler/broker/broker_config.go#L55
https://github.com/knative-sandbox/eventing-rabbitmq/blob/main/pkg/reconciler/broker/broker_config.go#L75

vaikas:
Well, we should still leave it in the backend in case somebody creates brokers without having installed the webhook, but still failing fast would be preferable.

n3wscott:
+1 do it, sounds awesome.

PR #309, opened 2021-05-25
Bring your own broker

vaikas:
/hold
Hold until I get tests for bring your own broker.

Split the reconciler into two pieces so that we can support bring your own broker without requiring that we install messaging-topology-operator & RabbitMQCluster operator since we can't use them with 'bring your own broker', and it seems silly to require them for that use case.

- :gift: Provide two versions of the controller, one that does not depend on rabbitmq cluster-operator nor messaging-topology-operator.
- :gift: One supports bring your own rabbitmq cluster
- :bug: If you use the cluster-operator / m-t-o version, we do not put finalizers on the brokers/triggers, which can lead to undeletable resources in some corner cases. 

/kind <kind>

Fixes #

**Release Note**

```release-note
 :gift: Provide an option for install that does not have dependency on rabbitmq cluster operator or messaging-topology-operator. That allows you to bring your own rabbitmq broker.
```

**Docs**

```docs

```


pkg/reconciler/brokerstandalone/controller.go
```diff
@@ -0,0 +1,139 @@
...
+limitations under the License.
+*/
+
+package brokerstandalone
+
+import (
+	"context"
+	"log"
+
+	"knative.dev/eventing-rabbitmq/pkg/client/injection/ducks/duck/v1beta1/rabbit"
```
n3wscott, line 23: nit: format

pkg/reconciler/brokerstandalone/resources/dispatcher.go
```diff
@@ -0,0 +1,123 @@
...
+const (
+	dispatcherContainerName = "dispatcher"
+)
+
+// DispatcherArgs are the arguments to create a Broker's Dispatcher Deployment that handles
+// DeadLetterSink deliveries.
+type DispatcherArgs struct {
+	Broker *eventingv1.Broker
+	Image  string
+	//ServiceAccountName string
```
n3wscott, line 43: delete?

pkg/reconciler/brokerstandalone/resources/dispatcher.go
```diff
@@ -0,0 +1,123 @@
...
+	RabbitMQHost       string
+	RabbitMQSecretName string
+	QueueName          string
+	BrokerUrlSecretKey string
+	BrokerIngressURL   *apis.URL
+	Subscriber         *apis.URL
+}
+
+func DispatcherName(brokerName string) string {
+	return fmt.Sprintf("%s-dlq-dispatcher", brokerName)
```
n3wscott, line 53: should this have something with the standalone name? maybe it is fine because you can't have two brokers of the same kind in the cluster of standalone vs not... ok I think it is ok

n3wscott (approved):

pkg/reconciler/broker/broker.go
```diff
@@ -145,45 +134,12 @@ func (r *Reconciler) ReconcileKind(ctx context.Context, b *eventingv1.Broker) pk
 		return err
 	}
 
-	if isUsingOperator(b) {
-		args.RabbitMQCluster = b.Spec.Config.Name
-		return r.reconcileUsingCRD(ctx, b, args)
-	} else {
-		return r.reconcileUsingLibraries(ctx, b, args)
+	if !isUsingOperator(b) {
+		// TODO: We need to flag this as an error. Or, maybe it comes from above (getExchangeArgs)
```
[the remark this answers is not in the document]
vaikas: I created an issue last week here:
https://github.com/knative-sandbox/eventing-rabbitmq/issues/304

If we do that, then we could just validate differently for both of these, so we only accept what we can support. I think it's better anyways so the resource fails fast instead of during runtime, wdyt?

config/brokerstandalone/500-broker-controller.yaml
```diff
@@ -0,0 +1,78 @@
+# Copyright 2020 The Knative Authors
```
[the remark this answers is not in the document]
vaikas: done.

n3wscott (approved):

vaikas:
/unhold
Should be good to go now.

n3wscott (approved):

PR #317, opened 2021-05-27
Added vhost to sources

### LINKED ISSUE #295: Add virtual host config support for rabbitmq source
kind/feature-request
 ## Overview
Add virtual host config support for rabbitmq source 
## Description
we need to config virtual host for this kind mq
image

gabo1208:
# Added vhost to RabbitMQSource definition

- 🎁 Added vhost optional property to RabbitMQSource definition and adapter

/kind feature-request

Fixes #295


failer.yaml
```diff
@@ -0,0 +1,31 @@
+# Copyright 2020 The Knative Authors
```
vaikas, line 1: Is this same as the file in config/failer/failer.yaml?

We probably don't want this at the root also.

pkg/adapter/adapter_test.go
```diff
@@ -450,3 +495,50 @@ func sinkAccepted(writer http.ResponseWriter, req *http.Request) {
 func sinkRejected(writer http.ResponseWriter, _ *http.Request) {
 	writer.WriteHeader(http.StatusRequestTimeout)
 }
+
+func TestAdapter_VhostHandler(t *testing.T) {
+	brokers := "amqp://localhost:5672"
```
vaikas, line 500: I think it would be more golang stylish to rewrite these tests as table driven tests.
Kind of like this:
https://github.com/knative-sandbox/eventing-rabbitmq/blob/main/pkg/reconciler/broker/resources/exchange_test.go#L20

so the struct and first two tests might look like this:
```
	for _, tt := range []struct {
		name      string
		brokers string
		vhost      string
		want       string
		}{{
		   name:      "no vhost",
		   brokers: "amqp://localhost:5672",
		   vhost: "",
		   want: "amqp://localhost:5672",
	  },{
		   name:      "vhost",
		   brokers: "amqp://localhost:5672",
		   vhost: "test-vhost",
		   want: "amqp://localhost:5672/test-vhost",
	}} {
	       t.Run(tt.name, func(t *testing.T) {
	             t.Parallel()
	             got := VhostHandler(tt.brokers, tt.vhost)
	             if got != tt.want {
	                   t.Errorf("Unexpected result, want %q got %q", tt.want, got)
	             }
                 })
           }

File: rabbitmq-broker.yaml [hunk 3]
```diff
@@ -0,0 +1,120 @@
+# Copyright 2020 The Knative Authors
```
vaikas, line 1: same here with the yaml file, looks like it's a combination of the files in the config/broker and we don't really want them in the root of the project.

pkg/adapter/adapter.go
```diff
@@ -104,10 +105,28 @@ func NewAdapter(ctx context.Context, processed adapter.EnvConfigAccessor, httpMe
 	}
 }
 
+func VhostHandler(Brokers string, Vhost string) string {
```
vaikas, line 108: If we don't expect to need to call this from any other place, maybe change this to:
vhostHandler

In go, a function starting with lowercase is not visible outside the package, so limiting to smallest scope possible is typically what you want.

third_party/VENDOR-LICENSE/LICENSE
```diff
@@ -0,0 +1,27 @@
+Copyright (c) 2009 The Go Authors. All rights reserved.
```
vaikas, line 1: Did you run ./hack/update-codegen.sh

These seem like leftover that should be removed by that script.

vaikas: Yeah, robot is unhappy :)
https://github.com/knative-sandbox/eventing-rabbitmq/pull/317/checks?check_run_id=2692258987

pkg/adapter/adapter_test.go
gabo1208: :o nice format, cool! fixed!

pkg/adapter/adapter.go
```diff
@@ -104,10 +105,28 @@ func NewAdapter(ctx context.Context, processed adapter.EnvConfigAccessor, httpMe
...
+	if len(Vhost) == 0 {
+		return Brokers
+	}
+
+	if Brokers[len(Brokers)-1] != '/' && Vhost[0] != '/' {
```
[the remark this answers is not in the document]
gabo1208: I pan to do today another PR with this change at least in this file and its tests

gabo1208:
Integrating new changes and simplifying some boilerplate issues with the PR

Issue #343, opened 2021-06-09
Unify the RabbitMQ resource naming into one place


vaikas:
Currently in both pkg/reconciler/{broker,trigger,brokerstandalone,triggerstandalone}/resources there are many duplicate functions dealing with names.

We should probably unify them all into something like:
pkg/rabbitnaming
and do it all in one place.

Also resolve this TODO while doing it (there's few others):
https://github.com/knative-sandbox/eventing-rabbitmq/blob/main/pkg/reconciler/broker/resources/exchange.go#L117

PR #341, opened 2021-06-08
Add support for Trigger DeadLetterSink for Standalone Broker.

vaikas:
:bug: Fix Standalone Broker to handle Trigger DeadLetterSink correctly.

-
-
-

/kind <kind>

Fixes #

**Release Note**

```release-note
:bug: Fix Standalone Broker to support Trigger DeadLetterSink properly.
```

**Docs**

```docs

```

PR #344, opened 2021-06-09
Rabbitmq resource naming

vaikas:
Fixes: #343 

Create sane names for RabbitMQ resources. Basically uses the UID of the controlling resource using standard kmeta.ChildName.

Builds on #341

-
-
-

/kind <kind>

Fixes #

**Release Note**

```release-note
 :broom: Use better names for all RabbitMQ resources. Both k8s resources as well as RabbitMQ resources for standalone Broker.
```

**Docs**

```docs

```


n3wscott (approved):
/lgtm
/approve

lints fixed in a followup

PR #356, opened 2021-07-01
vhost source e2e test structure and functionality

gabo1208:
# Changes

Added vhost feature e2e test


test/e2e/config/rabbitmqvhost/rabbitmqvhost.go
```diff
@@ -0,0 +1,37 @@
+/*
+Copyright 2020 The Knative Authors
```
n3wscott, line 2: nit: 2021

test/e2e/config/rabbitmqvhost/clustervhost.yaml
```diff
@@ -0,0 +1,28 @@
...
+kind: RabbitmqCluster
+metadata:
+  name: rabbitmqc
+  namespace: {{ .namespace }}
+spec:
+  rabbitmq:
+    additionalConfig: |
+      default_vhost = test-vhost
+      default_user = guest
+      default_pass = guest
```
n3wscott, line 25: is it safe to let a password be in the rabbitmq cluster CO? oh, this is not our code or resource. yikes, ok.

test/e2e/sourcevhost_test.go
```diff
@@ -0,0 +1,43 @@
...
+//
+// producer ---> rabbitmq --[vhost(source)]--> recorder
+//
+
+// VhostSourceTest makes sure an RabbitMQ Source is created on the desired vhost.
+func VHostSourceTest() *feature.Feature {
+	f := new(feature.Feature)
+
+	f.Setup("install RabbitMQ source on test-vhost", sourcevhost.Install())
+	f.Alpha("RabbitMQ source").Must("goes ready", AllGoReady)
```
n3wscott, line 40: This is a good smoketest, but we should follow-up or addition to this PR with a test that makes some events and we validate the source delivers the events, only then I think the passwords are known to be used?

Let me know what you want to do.

test/e2e/config/rabbitmqvhost/clustervhost.yaml
gabo1208: Dont know if oook XD but is an option that is usefull for this test purpose, ill try to improve this over the weekend like to try to refer the default user secret on the config, help is welcome :p

test/e2e/sourcevhost_test.go
gabo1208: mmm ok ok, i understand, definetly ill improve it with that :)

gabo1208: maybe a good follow up :)

test/e2e/sourcevhost_test.go
```diff
@@ -0,0 +1,52 @@
...
+
+//
+// producer ---> rabbitmq --[vhost(source)]--> recorder
+//
+
+// VhostSourceTest makes sure an RabbitMQ Source is created on the desired vhost.
+func VHostSourceTest() *feature.Feature {
+	f := new(feature.Feature)
+
+	f.Setup("install RabbitMQ source on test-vhost", sourcevhost.Install())
```
n3wscott, line 42: It is a little confusing to mix the producer with installing a vhost enabled broker.

I would break this into two steps: 

```go
	f.Setup("install RabbitMQ source on test-vhost", sourcevhost.Install())
	.Setup("install RabbitMQ source on test-vhost", vhostproducer.Install())
```

Then you are free to configure the producer independent of the source CO.

But we could make this a follow up issue and solve that problem when we get to it.

n3wscott (approved):
/lgtm

We can break apart the sender at a later time, there is even work in events hub to enable this for kafka so we can do the same here: https://github.com/knative-sandbox/reconciler-test/pull/225

PR #358, opened 2021-07-12
[WIP] Channel implementation

vaikas:
:gift: Add support for RabbitMQ channels.

-
-
-

/kind <kind>

Fixes #

**Release Note**

```release-note
:gift: Add support for RabbitMQ channels
```

**Docs**

```docs

```


isaac88:
Hello @benmoss 
Do you know why this issue has been closed ?
It's possible to use RabbitMQ as a Knative Eventing Channel ?
Thanks

gabo1208:
Hey @isaac88 , no, we don't have a rabbitmq channel implementation right now just broker and source

dpankros:
@gabo1208 Can I ask why?  I haven't been following this project for *that* long so there may be history I'm missing, but it seems like a huge hole for software that is considered GA.  Maybe I'm just not understanding the envisioned use-case for Rabbit with Knative.

We're trying to use Rabbit and Knative together and I was starting with a cron-based event bus.  Without a channel implementation I don't know how we could actually in good conscience use it in production for anything but the simplest of cases.  Just using a Parallel requires a channel implementation.  The InMemoryChannel is very clear that it's for dev use only and what's the other option? Implement Kafka so we can use Rabbit with Kafka channels?  Those are deprecated or Beta, anyway.  

This isn't just a Rabbit problem, BTW.  Knative doesn't list **any** Channel implementations that are not PoC, Beta, or Deprecated.  

Any insight would be appreciated.

gabo1208:
I think this is a great subject to get into Knative Eventing repo. But let me tell you a little bit about our Eventing RabbitMQ case (there should probably be a blog post or something explaining these things more clear than they are right now, I'll try to spread the word to the community about it).

Our specific case is, we wanted to have polished high level implementations that did everything we need in the integration with knative, basically Broker, Trigger and Sources are just enough to do everything related Knative Eventing. When you compare Channels, Subscriptions and Flows to Broker and Triggers, you see they do things in a lower level, what I mean by that is, let's take[ the Parallel example](https://github.com/knative/docs/blob/main/code-samples/eventing/parallel/multiple-branches/README.md): You need two services for filtering and the parallel creates a RabbitMQ channel underneath.
Event -> (Filter x2 -> Parallel (channel)) -> Transform x2

This can be done also by creating a Broker, sending the events, and filtering via Triggers from the broker itself:
Event ->(Broker -> Trigger (Filter) x2) -> Transform x2

So you just have to look the problem in a different way to achieve the Topology =)

And about the implementations, yep, let meping Eventing folks here to see what they have to say!

PR #360, opened 2021-07-14
Update Trigger behaviour for deadlettersink conditions due to upstream changes.

vaikas:
Add handling for Trigger deadlettersink condition that was updated in:
https://github.com/knative/eventing/pull/5551

The behaviour change can be seen in #354 so this also updates to latest deps so the tooling will start to work again.

-
-
-

/kind <kind>

Fixes #

**Release Note**

```release-note
🧹  Update the Trigger status DeadLetterSink condition to match that of upstream.
```

**Docs**

```docs

```


n3wscott (approved):
/lgtm
/approve

n3wscott:
/hold
Lint is complaining

vaikas:
Re: lint, yes, was due to deprecated methods, I'll fix in a followup 👍

PR #362, opened 2021-07-14
use source.StatsReporter from knative.dev/eventing

dprotaso:
## Proposed Changes

- :broom: we're shuffling the source.StatsReporter from knative.dev/pkg to knative.dev/eventing
  - Right now we setup an alias which will allow the transition to be done smoothly

**Release Note**

```release-note
NONE
```

**Docs**


vaikas:
lint is fixed in #361 
/lgtm
/approve

PR #369, opened 2021-07-23
Adapter fixes

### LINKED ISSUE #367: The RabbitMQ Source doesn't handle translating msgs to CloudEvents
Fixing those to-dos on:
- [1st](https://github.com/knative-sandbox/eventing-rabbitmq/blob/0852ca4c32d683ff7561a78e0d2a66904528c2ff/pkg/adapter/adapter.go#L312)
  - Doesn't re wrap messages that are already CloudEvents
  - Differentiate if the message is a CloudEvent in Binary or Structured encoding
  - Set all the required fields if the message is not a CloudEvent and have to be translated
- [2nd](https://github.com/knative-sandbox/eventing-rabbitmq/blob/0852ca4c32d683ff7561a78e0d2a66904528c2ff/pkg/adapter/adapter.go#L325)
  - Set CloudEvents content type from msg content type (if it is the case)

All of this is based on the CloudEvents RabbitMQ protocol binding specification ~~https://hackmd.io/VUONMvUkRH2103cbJFetdg~~ -> https://github.com/knative-sandbox/eventing-rabbitmq/blob/main/cloudevents-protocol-spec/spec.md

gabo1208:
Fixes #367

### DIFF
--- a/pkg/adapter/adapter.go
+++ b/pkg/adapter/adapter.go
@@ -359,7 +359,8 @@ func setEventContent(a *Adapter, msg wabbit.Delivery, contentType string) (cloud
 }
 
 func setEventBatchContent(a *Adapter, msg wabbit.Delivery) ([]cloudevent.Event, error) {
-	// Needs a batch size
+	// Would be good to have a batch size parameter
+	// Process all the events, so it does not send any events
 	var payload []cloudevent.Event
 	err := json.Unmarshal(msg.Body(), &payload)
 	if err != nil {
--- a/pkg/adapter/adapter.go
+++ b/pkg/adapter/adapter.go
@@ -304,15 +304,17 @@ func (a *Adapter) PollForMessages(channel *wabbit.Channel,
 	}
 }
 
-func setMsgContentType(msg wabbit.Delivery) string {
+func msgContentType(msg wabbit.Delivery) string {
 	contentType := fmt.Sprint(msg.Headers()["Content-type"])
 	switch contentType {
-	case cloudevent.ApplicationJSON, cloudevent.TextJSON, "":
+	case cloudevent.ApplicationJSON, cloudevent.TextJSON:
 		return cloudevent.ApplicationJSON
 	case cloudevent.ApplicationCloudEventsJSON, cloudevent.ApplicationCloudEventsBatchJSON:
 		return cloudevent.ApplicationCloudEventsJSON
-	default:
+	case cloudevent.ApplicationXML, cloudevent.Base64, cloudevent.TextPlain:
 		return contentType
+	default:
+		return cloudevent.ApplicationJSON
 	}
 }
 
--- a/pkg/adapter/adapter.go
+++ b/pkg/adapter/adapter.go
@@ -306,11 +306,11 @@ func (a *Adapter) PollForMessages(channel *wabbit.Channel,
 	}
 }
 
-func msgContentType(msg *wabbit.Delivery) (string, bool, error) {
+func msgContentType(msg wabbit.Delivery) (string, bool, error) {
 	var contentType string
 	var isBinary bool
 
-	for key, val := range (*msg).Headers() {
+	for key, val := range msg.Headers() {
 		switch strings.ToLower(key) {
 		case "content-type":
 			if s, ok := val.(string); ok {
@@ -332,31 +332,31 @@ func msgContentType(msg *wabbit.Delivery) (string, bool, error) {
 	return contentType, isBinary, nil
 }
 
-func setEventContent(a *Adapter, msg *wabbit.Delivery, contentType string) (cloudevent.Event, error) {
+func setEventContent(a *Adapter, msg wabbit.Delivery, contentType string) (cloudevent.Event, error) {
 	event := cloudevents.NewEvent()
 	if contentType == cloudevent.ApplicationCloudEventsJSON {
-		err := json.Unmarshal((*msg).Body(), &event)
+		err := json.Unmarshal(msg.Body(), &event)
 		if err != nil {
 			return event, err
 		}
 	} else {
-		if (*msg).MessageId() != "" {
-			event.SetID((*msg).MessageId())
+		if msg.MessageId() != "" {
+			event.SetID(msg.MessageId())
 		} else {
 			event.SetID(string(uuid.NewUUID()))
 		}
 
-		event.SetTime((*msg).Timestamp())
+		event.SetTime(msg.Timestamp())
 		event.SetType(sourcesv1alpha1.RabbitmqEventType)
 		event.SetSource(sourcesv1alpha1.RabbitmqEventSource(
 			a.config.Namespace,
 			a.config.Name,
 			a.config.Topic,
 		))
-		event.SetSubject((*msg).MessageId())
-		event.SetExtension("key", (*msg).MessageId())
+		event.SetSubject(msg.MessageId())
+		event.SetExtension("key", msg.MessageId())
 
-		err := event.SetData(contentType, (*msg).Body())
+		err := event.SetData(contentType, msg.Body())
 		if err != nil {
 			return event, err
 		}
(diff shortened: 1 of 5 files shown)


gabo1208:
Im finishing the tests currently and i have some questions that ill leave as comments in the appropiated lines

pkg/adapter/adapter.go
```diff
@@ -302,30 +304,52 @@ func (a *Adapter) PollForMessages(channel *wabbit.Channel,
 	}
 }
 
+func setMsgContentType(msg wabbit.Delivery) string {
+	contentType := fmt.Sprint(msg.Headers()["Content-type"])
```
gabo1208, line 308: its ok to use this or should i search for the datacontenttype header as well?

pkg/adapter/adapter.go
```diff
@@ -302,30 +304,52 @@ func (a *Adapter) PollForMessages(channel *wabbit.Channel,
...
+	switch contentType {
+	case cloudevent.ApplicationJSON, cloudevent.TextJSON, "":
```
gabo1208, line 310: I used just the types that were defined on the event library

pkg/adapter/adapter.go
vaikas: I do not think it should matter for here since we're not really consuming the event. It should only be for the consumer that can decide how to decode the actual data.
https://github.com/cloudevents/spec/blob/v1.0.1/spec.md#datacontenttype

vaikas: FYI, I had been wanting to use this protocol adapter here:
https://github.com/cloudevents/sdk-go/tree/82f2b61ecde41fd0577969cc58a7c2a18eeda250/protocol/amqp/v2

But it's for different version than what the rabbitmq supports. However, there's probably some interesting learnings there as well.

vaikas: FWIW, here's the spec for amqp 1.0 messaging formats for the CloudEvents.
https://github.com/cloudevents/spec/blob/v1.0.1/amqp-protocol-binding.md

pkg/adapter/adapter.go
```diff
@@ -302,30 +304,52 @@ func (a *Adapter) PollForMessages(channel *wabbit.Channel,
...
+		return cloudevent.ApplicationJSON
+	case cloudevent.ApplicationCloudEventsJSON, cloudevent.ApplicationCloudEventsBatchJSON:
+		return cloudevent.ApplicationCloudEventsJSON
+	default:
+		return contentType
```
vaikas, line 315: I wonder if this should return an error if there's no contenttype?

pkg/adapter/adapter.go
```diff
@@ -302,56 +305,127 @@ func (a *Adapter) PollForMessages(channel *wabbit.Channel,
...
+
+	for key, val := range msg.Headers() {
+		if strings.ToLower(key) == "content-type" {
+			contentType = strings.ToLower(fmt.Sprint(val))
+			break
+		}
+	}
+
+	if contentType == "" {
+		return "", fmt.Errorf("no content type present on Rabbitmq msg")
```
vaikas, line 319: I think instead of fmt.Errorf you might prefer to just do errors.New since there's no formatting done here.

pkg/adapter/adapter.go
```diff
@@ -302,56 +305,127 @@ func (a *Adapter) PollForMessages(channel *wabbit.Channel,
...
-	a.logger.Info("url ->" + a.httpMessageSender.Target)
-	req, err := a.httpMessageSender.NewCloudEventRequest(a.context)
-	if err != nil {
-		return err
+func msgContentType(msg wabbit.Delivery) (string, error) {
+	var contentType string
+
+	for key, val := range msg.Headers() {
+		if strings.ToLower(key) == "content-type" {
+			contentType = strings.ToLower(fmt.Sprint(val))
```
vaikas, line 313: instead of fmt.Sprint, I wonder if we should be doing the type assertion, or at least reading here I'd think that's what we should be doing?
https://github.com/streadway/amqp/blob/master/types.go#L221

pkg/adapter/adapter.go
```diff
@@ -302,56 +305,127 @@ func (a *Adapter) PollForMessages(channel *wabbit.Channel,
...
-		return err
+		return payload, err
 	}
 
-	err = http.WriteRequest(a.context, binding.ToMessage(&event), req)
+	return payload, nil
+}
+
+func (a *Adapter) postMessage(msg wabbit.Delivery) error {
+	a.logger.Info("url ->" + a.httpMessageSender.Target)
```
vaikas, line 376: Maybe debug instead of Info? This is going to log for every event, right?

pkg/adapter/adapter.go
```diff
@@ -302,56 +305,127 @@ func (a *Adapter) PollForMessages(channel *wabbit.Channel,
...
+	req, err := a.httpMessageSender.NewCloudEventRequest(a.context)
 	if err != nil {
 		return err
 	}
 
-	res, err := a.httpMessageSender.Send(req)
+	a.logger.Info(fmt.Sprint(msg.Headers()))
```
vaikas, line 382: probably do not want to log all the headers in case they contain something sensitive?

test/e2e/cmd/rabbitproducer/main.go
```diff
@@ -77,6 +77,7 @@ func main() {
 			false,  // mandatory
 			false,  // immediate
 			amqp.Publishing{
+				Headers:     amqp.Table{"content-type": "text/plain"},
```
vaikas, line 80: So this means that for every msg in the rabbitmq, we require content-type? Is that ok? Just curious if that would then mean that folks would have to change the messages they produce? Or put another way, is there a way to assume that if there's none, we would treat missing as text/plain?
@n3wscott since he also has looked at Rabbit.

pkg/adapter/adapter.go
gabo1208: that was just me testing things and forgetting to delete them hehe :p

pkg/adapter/adapter.go
gabo1208: This one is just for one event

test/e2e/cmd/rabbitproducer/main.go
gabo1208: Yup i think i need a way to get the content type property, but i cant find how to do it in the wabbit delivery interface. Need a little help with that

pkg/adapter/adapter.go
gabo1208: One per event or per batch*

gabo1208: It seems so, I didn't put that piece of code there

pkg/adapter/adapter_test.go
```diff
@@ -689,6 +704,10 @@ func TestAdapter_msgContentType(t *testing.T) {
 			if tt.err && err == nil {
 				t.Errorf("Unexpected error state for msg type want:\n%+s\ngot:\n%+s", tt.want, got)
 			}
+
```
vaikas, line 707: I wonder if the L700 above should be:

```
if err == nil && got != tt.want {
```
When there's no error we should check if got == want or am I missing something?

gabo1208: Here im checking if there is an error, but yeah the error shouldnt have the tt.want but the error message let me change that :)

vaikas: I'm saying that L700 checks if got != tt.want only if there's an error? That seems wrong? msgContentType returns err == nil if all goes well, and that's when we should be testing if the got == tt.want, not when there's an error like what we're doing here?

pkg/adapter/adapter.go
n3wscott: those are just the minimum types, in the sdk you can register anything. We should expect everything and anything to be enqueued.

PR #375, opened 2021-08-16
Address a long-overdue TODO

mattmoor:
/kind cleanup

**Release Note**

```release-note

```

**Docs**

```docs

```


mattmoor:
cc @dprotaso (who originally removed a lot of the K8s pinning)

mattmoor:
```
coding error: you must register resource to list kind for every resource you're going to LIST when creating the client.  See NewSimpleDynamicClientWithCustomListKinds or register the list into the scheme: rabbitmq.com/v1beta1, Resource=rabbitmqclusters out of map[...
```

Looks like the `rabbitmqclusters` type isn't being registered with the scheme, so things blow up with the newer K8s dynamic client.

mattmoor:
More specifically, the list type (in this panic).

PR #378, opened 2021-08-17
Rework the way eventing-rabbitmq does Rabbit codegen

mattmoor:
Today we rerun codegen for the Rabbit clients to get injection, but because they use kubebuilder their layout is a bit strange.  There's a hack to temporarily copy the api types into the right structure, and then undo it, which is a bit fragile (if codegen references that package, then it won't build!).  We also try to contort so that we can use their existing client/informer/lister codegen, which isn't really necessary.

With this change, we instead permanently copy the types into a suitable structure under `third_party/pkg/apis` and generate all of the clients we need under `third_party/pkg/client`.  This is a slight superset of what Istio and Contour are already doing (we don't need to copy those types), and feels a bit simpler (and less fragile) than what we are currently doing.

/kind cleanup
/assign @n3wscott 

**Release Note**

```release-note

```

**Docs**

```docs

```


n3wscott:
This seems like a fine way to handle out of tree codegen for oddly layed out apis. 

👍

PR #399, opened 2021-09-02
Bump topology-operator to 1.0.0

benmoss:
# Changes

- :broom: Update rabbitmq-topology-operator to 1.0.0

/kind cleanup


benmoss:
Looks like we might be stuck until eventing bumps to client-go 0.22

PR #402, opened 2021-09-03
drop scheduled actions these fill up the job queue causing delays

dprotaso:


benmoss:
/lgtm
/approve

thanks @dprotaso !

Issue #409, opened 2021-09-08
Be smarter about retries


embano1:
The dispatcher currently considers all !2xx || !3xx status codes as retryable, which is somewhat against the HTTP protocol spec:

https://github.com/knative-sandbox/eventing-rabbitmq/blob/928976461374266e0c029bc61719c68b3dffad8b/pkg/dispatcher/dispatcher.go#L160

I'm happy to file a PR if we get consensus on which codes we want to retry. Related work:

- https://docs.aws.amazon.com/apigateway/api-reference/handling-errors/
- https://github.com/hashicorp/go-retryablehttp/blob/master/client.go#L424

xtreme-sameer-vohra:
Thanks for filing this @embano1 
Seems pretty reasonable to be more specific. At minimum we should cover the explicit http status checks.

benmoss:
https://github.com/knative/specs/blob/main/specs/eventing/data-plane.md#event-acknowledgement-and-delivery-retry

benmoss:
https://github.com/knative-sandbox/eventing-kafka-broker/blob/3b860bf2870caa2b27e39683d7507fd6dd6cd136/data-plane/dispatcher/src/main/java/dev/knative/eventing/kafka/broker/dispatcher/impl/WebClientCloudEventSender.java#L84-L90

https://github.com/knative/eventing/blob/f031ba23b23d20ed8d13491000feb0abcf50d562/pkg/channel/message_dispatcher.go#L312-L315

It seems this might be the standard among implementations right now

embano1:
I wonder why this is overwritten at all and not the default option used? https://github.com/cloudevents/sdk-go/blob/c6de9589bb3578364689adc64414000dabe93c01/v2/protocol/http/protocol.go#L40

This looks like it does what we need (although retrying on 500 would be nice too, I may want to file a PR for the SDK)?

benmoss:
Yeah, it definitely seems like something the sdk should be handling

lionelvillard:
just in case you haven't seen this: https://github.com/knative/eventing/blob/main/pkg/kncloudevents/retries.go

and this: https://github.com/knative/eventing/pull/5285

embano1:
@lionelvillard are you suggesting we should be using kncloudevents as client?

lionelvillard:
I'm just raising awareness, not suggesting anything :-) kncloudevents is meant to be Knative eventing compliant so maybe it makes sense for you to use it.

[...]

PR #276, opened 2021-04-21
grab the latest with test resource cleanups

vaikas:
# Changes

- Grab the latest version of eventing. Also add an import to eventing event-library test image import that was breaking the conformance tests from being built.
- :bug: Fix a bug in which trigger delivery was not used. Fixes: #219 
-

/kind <kind>

Fixes #

**Release Note**

```release-note

```

**Docs**

```docs
:bug: Fix a bug in which trigger delivery spec was not being used, was always only using Brokers one.
```

PR #418, opened 2021-09-15
Add support for concurrent dispatch and configurable prefetch

### LINKED ISSUE #329: [feature request] allow set prefetch on dispatcher
right now, the prefetch on dispatcher is set to 1. This seems to be very restrictive, especially for long-running consumers.

I'm wondering why 1 is chosen?

event with keda to scale it based on the number of messages, it still seems inefficient.

Is there a plan to allow set the prefetch count or increase the default?

thx!

benmoss:
# Changes

- :gift: Add ability to configure the dispatcher's prefetch
- :gift: Dispatcher sends events concurrently

/kind enhancement

Fixes #329

**Release Note**

```release-note
:page_facing_up: A trigger will now send events to its subscriber in parallel, with the number of in-flight messages defaulting to 10 and configurable via the annotation`rabbitmq.eventing.knative.dev/prefetchCount`.
```

/assign @embano1 @gabo1208 

Users can now set the annotation`rabbitmq.eventing.knative.dev/prefetchCount` to control the concurrency of the dispatcher and the number of in-flight events allowed. This has the implication that event order is no longer guaranteed with prefetch counts > 1. The default prefetch if not specified is 10, this is totally arbitrary and I'm open to opinions on a more sensible default if anyone has any 😄 .


pkg/dispatcher/dispatcher.go
```diff
@@ -96,63 +96,7 @@ func (d *Dispatcher) ConsumeFromQueue(ctx context.Context, channel wabbit.Channe
...
-					}
-					continue
-				}
-			}
-
-			err = msg.Ack(ackMultiple)
-			if err != nil {
-				logging.FromContext(ctx).Warn("failed to ACK event: ", err)
-			}
+			go d.dispatch(ctx, msg, ceClient)
```
embano1, line 99: This could lead to a situation where, if ctx is canceled, we don't wait for this goroutine to return cleanly. Not sure what the contract should be here since the scenario also depends on how long the goroutine (sink) executes (could be secs/mins).

embano1 (commented):
Questions:

1) Regarding this comment: https://github.com/knative-sandbox/eventing-rabbitmq/blob/c70eab331eeb664526a2b54c6e630f33c7938789/pkg/dispatcher/dispatcher.go#L144 -> IMHO the SDK won't retry at all if set to 0.

1) https://github.com/knative-sandbox/eventing-rabbitmq/blob/c70eab331eeb664526a2b54c6e630f33c7938789/pkg/dispatcher/dispatcher.go#L166 < this is also available as env, so use this instead?

1) https://github.com/knative-sandbox/eventing-rabbitmq/blob/c70eab331eeb664526a2b54c6e630f33c7938789/pkg/dispatcher/dispatcher.go#L168 < use retries defined in env?

1) Should we define a custom consumer name (ie dispatcher) to simplify troubleshooting (especially when adding auto-scaling later)? https://github.com/knative-sandbox/eventing-rabbitmq/blob/c70eab331eeb664526a2b54c6e630f33c7938789/pkg/dispatcher/dispatcher.go#L68

1) Do you also want to tackle https://github.com/knative-sandbox/eventing-rabbitmq/issues/409 in this PR?

cmd/dispatcher/main.go
```diff
@@ -41,14 +41,15 @@ type envConfig struct {
...
 
+	// Number of concurrent messages in flight
+	PrefetchCount int           `envconfig:"PREFETCH_COUNT" default:"10" required:"false"`
 	Retry         int           `envconfig:"RETRY" required:"false"`
 	BackoffPolicy string        `envconfig:"BACKOFF_POLICY" required:"false"`
 	BackoffDelay  time.Duration `envconfig:"BACKOFF_DELAY" required:"false"`
 }
 
 const (
 	defaultBackoffDelay = 50 * time.Millisecond
```
gerhard, line 52: I would set this as a `BackoffDelay` default.

cmd/dispatcher/main.go
```diff
@@ -41,14 +41,15 @@ type envConfig struct {
...
-	defaultPrefetch     = 1
 	defaultPrefetchSize = 0
```
gerhard, line 53: I would remove this constant and set it directly in the `channel.Qos/3` function call.

gerhard (commented):
A prefetch of 10 is a decent default, and better than 1 if consumer message ordering is not needed. This portion of the RabbitMQ docs explain it well: [Consumer Acknowledgement Modes, Prefetch and Throughput](https://www.rabbitmq.com/confirms.html#channel-qos-prefetch-throughput). While a value of 100 or more can result in higher throughput, this tends to be workload specific (e.g. message payload, number of consumers etc.) and it can result in the broker using more memory than necessary. Messages that have been delivered and are waiting for ACKs will be kept in memory not just on the client side, but also on the broker side. So yes, 10 is a decent prefetch default.

pkg/dispatcher/dispatcher.go
benmoss: If the context is cancelled then the requests will be cancelled as well. I updated the implementation to handle waiting on the goroutines to finish.

cmd/dispatcher/main.go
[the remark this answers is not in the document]
benmoss: Where do you see that we force no-ack? https://www.rabbitmq.com/amqp-0-9-1-reference.html#domain.no-ack

That would imply we don't use acknowledgements, which is not true

cmd/dispatcher/main.go
benmoss: Not sure what you mean. The prefetch count in milliseconds should be the backoff delay?

benmoss:
> Questions:
> 
> 2. https://github.com/knative-sandbox/eventing-rabbitmq/blob/c70eab331eeb664526a2b54c6e630f33c7938789/pkg/dispatcher/dispatcher.go#L166
>     < this is also available as env, so use this instead?

It looks like this used to be this way, and then @vaikas changed it in https://github.com/knative-sandbox/eventing-rabbitmq/pull/276/files#diff-7493504960c3b7e5e2f7c2d0a09116c08af37ebcd3798dced0c4478781443c3aR114-R121

Not sure I understand the comment here, "cloudevents is the max total tries".

> 3. https://github.com/knative-sandbox/eventing-rabbitmq/blob/c70eab331eeb664526a2b54c6e630f33c7938789/pkg/dispatcher/dispatcher.go#L168
>     < use retries defined in env?

Same

> 4. Should we define a custom consumer name (ie dispatcher) to simplify troubleshooting (especially when adding auto-scaling later)? https://github.com/knative-sandbox/eventing-rabbitmq/blob/c70eab331eeb664526a2b54c6e630f33c7938789/pkg/dispatcher/dispatcher.go#L68

This gets autogenerated right now, 
https://github.com/knative-sandbox/eventing-rabbitmq/blob/concurrency/vendor/github.com/streadway/amqp/consumers.go#L19-L21

On my deployment this comes out to `ctag-/ko-app/dispatcher-1`, seems okay to me.

> 5. Do you also want to tackle #409 in this PR?
Let's keep this a separate issue

cmd/dispatcher/main.go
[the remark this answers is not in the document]
embano1: This is how I'm reading go-amqp and the Rabbit docs:

> When autoAck (also known as noAck) is true, the server will acknowledge
deliveries to this consumer prior to writing the delivery to the network.  When
autoAck is true, the consumer should not call Delivery.Ack.

https://github.com/knative-sandbox/eventing-rabbitmq/blob/c944a0c5a3bea768ce200c6812d44072b4bb7001/vendor/github.com/streadway/amqp/channel.go#L1017

> The prefetch-size is ignored if the no-ack option is set.

https://www.rabbitmq.com/amqp-0-9-1-reference.html#basic.qos.prefetch-size

We set `autoAck=false`: https://github.com/knative-sandbox/eventing-rabbitmq/blob/c944a0c5a3bea768ce200c6812d44072b4bb7001/pkg/dispatcher/dispatcher.go#L70

pkg/dispatcher/dispatcher.go
embano1: Thx, I was referring to waiting for goroutines 👍

embano1:
Sorry for my confusing comments, let me put them here one by one (since they're not part of your PR but warrant some discussion here IMHO):

This comment should be safe to drop: https://github.com/knative-sandbox/eventing-rabbitmq/blob/c70eab331eeb664526a2b54c6e630f33c7938789/pkg/dispatcher/dispatcher.go#L144

embano1:
Use env `BACKOFF_DELAY` ([1]) instead of hardcoded value here? https://github.com/knative-sandbox/eventing-rabbitmq/blob/c70eab331eeb664526a2b54c6e630f33c7938789/pkg/dispatcher/dispatcher.go#L166

[1]
https://github.com/knative-sandbox/eventing-rabbitmq/blob/c70eab331eeb664526a2b54c6e630f33c7938789/cmd/dispatcher/main.go#L48

pkg/dispatcher/dispatcher.go
```diff
@@ -172,10 +118,73 @@ func isSuccess(ctx context.Context, result protocol.Result) bool {
...
+			}
+			continue
+		}
+
+		logging.FromContext(ctx).Debugf("Got event as: %+v", event)
+		ctx = cloudevents.ContextWithTarget(ctx, d.SubscriberURL)
+
+		// Our dispatcher uses Retries, but cloudevents is the max total tries. So we need
+		// to adjust to initial + retries.
+		// TODO: What happens if I specify 0 to cloudevents. Does it not even retry.
```
embano1, line 149: drop this comment?

pkg/dispatcher/dispatcher.go
```diff
@@ -172,10 +118,73 @@ func isSuccess(ctx context.Context, result protocol.Result) bool {
...
+				logging.FromContext(ctx).Warn("failed to NACK event: ", err)
+			}
+			continue
+		}
+
+		logging.FromContext(ctx).Debugf("Got Response: %+v", response)
+		if response != nil {
+			logging.FromContext(ctx).Infof("Sending an event: %+v", response)
+			ctx = cloudevents.ContextWithTarget(ctx, d.BrokerIngressURL)
+			backoffDelay := 50 * time.Millisecond
```
embano1, line 171: Use value from env `BACKOFF_DELAY` here as well?

pkg/dispatcher/dispatcher.go
```diff
@@ -172,10 +118,73 @@ func isSuccess(ctx context.Context, result protocol.Result) bool {
...
+			// Use the retries so we can just parse out the results in a common way.
+			cloudevents.ContextWithRetriesExponentialBackoff(ctx, backoffDelay, 1)
```
embano1, line 173: hardcoded retries to one or use env defined retries?

embano1 (commented):
> Not sure I understand the comment here

Sorry for the confusion, I left my comments inline. Looks good overall to me!

cmd/dispatcher/main.go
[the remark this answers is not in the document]
benmoss: Oh, I think this is bad documentation on rabbit's part. I think they mean 

> The prefetch-size is ignored if the no-ack option is set *to true*.

Since we set it to false, we are not using autoAck/noAck, and therefore prefetch size/prefetch count is still relevant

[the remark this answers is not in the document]
benmoss: @gerhard can you back me up on this interpretation?

pkg/reconciler/trigger/resources/dispatcher_test.go
```diff
@@ -257,78 +191,36 @@ func TestMakeDLXDispatcherDeployment(t *testing.T) {
...
+		} else {
+			args.Trigger.ObjectMeta.Annotations[prefetchAnnotation] = c
+		}
 	}
 }
+
+func Int32Ptr(i int32) *int32 {
+	return &i
+}
+func StringPtr(s string) *string {
```
gabo1208, line 224: anything planned for this stringPtr function? is not been used

benmoss: removed, thanks!

salaboy:
@benmoss after reviewing this PR with my morning coffee at hand I can see that it is doing what we discussed yesterday. 
This PR with a page in the documentation showing how to use the label and the topology created on the rabbitMQ side will be a great addition.

gvmw:
💯

JamesMcMahon:
Is there documentation around how to use this feature?

For instance, is this correct?

```yaml
kind: RabbitmqSource
metadata:
  name: rabbitmq-source
  annotations:
    rabbitmq.eventing.knative.dev/prefetchCount: "10000"
# ...
# additional config below
```

embano1:
@JamesMcMahon IMHO the annotation is only respected on a `trigger` resource

PR #432, opened 2021-09-23
Use latest versions in operator docs

gvmw:
# Changes

- 📚 Update or clean up installation docs

/kind documentation

Rather than using specific versions, and always needing to update these docs (or have them go outdated), use `latest` instead.

While at it, I did some minor cleanup and also was wondering about the abrupt ending. How should we continue this documentation page? I've proposed a few ideas under **Next steps** - WDYT?

### DIFF
--- a/broker/operator-based.md
+++ b/broker/operator-based.md
@@ -28,7 +28,7 @@
 
 ## Installation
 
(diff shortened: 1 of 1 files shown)


gvmw:
I have added the next step which worked in our (+@ikvmw) experience, this is now ready for review.

gabo1208 (approved):

PR #441, opened 2021-09-30
Simplify backoff delay default

gvmw:
# Changes

- 🧹 Update or clean up current behavior

/kind cleanup

Follow-up to https://github.com/knative-sandbox/eventing-rabbitmq/pull/418/files#r713360735

PR #446, opened 2021-10-04
Fix BackoffDelay default

gvmw:
# Changes

- 🐛  Fix default value for BackOff Delay

/kind bug

Fix  for https://github.com/knative-sandbox/eventing-rabbitmq/pull/441

Must include time unit, otherwise time.Duration type will fail.


gvmw:
cc @benmoss

PR #486, opened 2021-11-02
Use retry logic from kncloudevents

### LINKED ISSUE #409: Be smarter about retries

benmoss:
This matches the behavior in the spec:
https://github.com/knative/specs/blob/cfc57cb5a115e026c890044f591241c96087923f/specs/eventing/data-plane.md#event-acknowledgement-and-delivery-retry

# Changes

- :bug: Fix dispatcher retry logic to match the eventing spec

/kind bug

Fixes #409

https://github.com/knative/eventing/blob/1e4fa3571ad7f08ddae9a7558aee3b508334e3f4/pkg/kncloudevents/retries.go#L123-L131

I don't really know what the `-1` bit is about.

**Release Note**

```release-note
:page_facing_up: The trigger dispatcher will now only retry deliveries when response status codes status are 5XX, 404, 409, 429, or -1.
```

### DIFF
--- a/pkg/dispatcher/dispatcher.go
+++ b/pkg/dispatcher/dispatcher.go
@@ -19,6 +19,7 @@ package dispatcher
 import (
 	"context"
 	"encoding/json"
+	"net/http"
 	"sync"
 	"time"
 
@@ -30,6 +31,7 @@ import (
 	amqperr "github.com/rabbitmq/amqp091-go"
 	"go.uber.org/zap"
 	eventingduckv1 "knative.dev/eventing/pkg/apis/duck/v1"
+	"knative.dev/eventing/pkg/kncloudevents"
 	"knative.dev/pkg/logging"
 )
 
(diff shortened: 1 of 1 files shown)


gabo1208 (approved):

embano1:
Follow-up question: the `ceclient` is set up with `ceClient, err := cloudevents.NewClientHTTP(cehttp.WithIsRetriableFunc(isRetriableFunc))`. This func has different retry code logic than `isSuccess()` which uses `kncloudevents.SelectiveRetry`.

```go
func isRetriableFunc(sc int) bool {
	return sc < 200 || sc >= 300
}
```

Would it make sense to drop `isRetriableFunc` in favor of `kncloudevents.SelectiveRetry`?

PR #445, opened 2021-10-04
[WIP] Add RMQ broker performance test

ikvmw:
see #239

# Changes

-
-
-

/kind <kind>

Fixes #

**Release Note**

```release-note

```

**Docs**

```docs

```

Issue #334, opened 2021-06-04
Ensure Broker Ingress uses Publisher-Confirm.


vaikas:
The channel must be configured to use Publisher Confirm, since the acks are asynchronous. As per this article here:
https://www.rabbitmq.com/tutorials/tutorial-seven-java.html#:~:text=Publisher%20confirms%20are%20a%20RabbitMQ,of%20on%20the%20server%20side.

MarcialRosales:
@gvmw Indeed it is a good way to exert back-pressure, but it is also the only way to ensure guarantee of delivery of messages for senders. Otherwise, it is a fire-and-forget model which is not what the majority of customers expect. 

As you propose, the simplest solution is to force a FIFO model by blocking further HTTP requests until the previous send has received its publisher-confirm. But I guess you can configure the HTTP Sink to accept up to N outstanding HTTP request (or send operations). That way you have extra concurrency and yet ensure delivery.

ikvmw:
This is what I have in my head so far:

## RabbitMQ
1. The main goal for publisher-confirms is to make publishing reliable. That is it: give publisher good amount of guarantees. For example if message is a persistent message it will be acked only after storing on disk (note the expected latency - "hundreds of milliseconds" as per https://www.rabbitmq.com/confirms.html#publisher-confirms-latency)
2. As a side-effect publisher-confirms can be used to throttle publishers. Again, persistent message are the obvious example here - if disk is overloaded the confirms will arrive later and iff publisher has some sort of a limit regarding unacknowledged messages count, at some point it will have to stop publishing and wait.
3. In the RabbitMQ world publisher-confirms are opt-in and can be used only for a subset of messages but a publisher. Even inside same connection only some channels can be in confirm mode.

## Knative
Knative wants to help its users to focus on important things. Providing good defaults, safety, flexibility and so on.
It will be good to not overload broker.

## On safety and Knative guarantees. 
We ca use publisher-confirms to make publishing synchronous and say to users that if sink returned 200 broker accepted the event

I personally don't see much value in this yet, since having routable message doesn't imply or guarantee something useful will come out. The receiving end can be down or not existing or overloaded and so on. In my view apps should be ready to process lost events regardless of the reason.

## Using publisher-confirms for throttling and safety

- The simplest way - make all requests synchronous inside sink instance by using mutex. Do not process next HTTP request till this one not acked. I think it kills parallelism.
[...]

PR #491, opened 2021-11-04
Update performance benchmarks post 1.0.0 release

### LINKED ISSUE #412: How can I benchmark eventing-rabbitmq?
Blocked on #239

## Ask
Instructions to setup and validate requirements listed under #239 - `Expectation from Developer perspective` given 
- a fresh K8s cluster > 1.21.0
- KUBECONFIG pointing to cluster

## References
Eventing - https://github.com/knative/eventing/tree/main/test/performance
Kafka - https://github.com/knative-sandbox/eventing-kafka/tree/main/test/performance

gvmw:
# Changes

- :broom: Update performance benchmark

/kind cleanup

Closes #412 for good. This is a follow-up to https://github.com/knative-sandbox/eventing-rabbitmq/pull/445.

I have confirmed that everything works as expected. I have removed all "smart" scripting and went with the plain approach. I have also included the results of the benchmarks in this repository, rather than referencing files that were uploaded as a GitHub thread comment.

There are a few unexpected behaviours here which I have captured in the README. Expanding on them below:

In previous iterations we didn't have any Receiver throughput values to plot. I am not sure why, but I have confirmed that we now have them.

When plotting throughput, especially receiver throughput, gnuplot can now take up to 5 minutes to load the graph. I suspect that this is due to the solid area which gets rendered. We never had it before, so this is a new & unexpected behaviour.

Screenshot 2021-11-04 at 13 00 25

When publishing 1k messages per second, the receiver cannot keep up. Increasing the CPU count from 1 to 2 doesn't help. I have attached a screenshot in the PR which captures the expected performance. We may want to troubleshoot, but for now we should just acknowledge that the 1k msg/s performance test may work for short bursts, but the workload is unsustainable (producers are constantly outpacing consumers).

Screenshot 2021-11-04 at 13 17 50

By the way, this 1k msg/s benchmark will take a really long time to complete (didn't wait), because the receiver consumes 3.4 msg/s 🤨

image

Lastly, I have confirmed that https://github.com/knative-sandbox/eventing-rabbitmq/issues/334#issuecomment-942411656 is spot-on from the broker perspective. Without it, the producer never waits for the broker and simply keeps pushing messages, like UDP does (fire & forget).


gabo1208:
I'll wait for @xtreme-sameer-vohra and @benmoss to see this for the merge

xtreme-sameer-vohra:
> By the way, this 1k msg/s benchmark will take a really long time to complete (didn't wait), because the receiver consumes 3.4 msg/s 🤨

The results between the 2 benchmarks original(135K in 3 mins) vs 1K Sustained(300K in 5 mins) seem **very** different 🤔 based on graphs. What am I missing ?

gvmw:
> The results between the 2 benchmarks original(135K in 3 mins) vs 1K Sustained(300K in 5 mins) seem very different 🤔  based on graphs. What am I missing ?

The original benchmark is 500..1000 msg/s for 30 seconds in steps of 100, so 500 - 600 - 700 - 800 - 900 - 1000.

The last benchmark is 1000 msg/s sustained for 300 seconds, which is a struggle because the message backlog keeps growing and everything slows down to a crawl (3.5 msg/s).

Do the new instructions work for you @xtreme-sameer-vohra? I think that's what we care about in the context of this PR. Yes, we should look into the 1000 msg/s scenario, but as a separate PR / issue I think 🙂

PR #493, opened 2021-11-05
Add predeclared behavior in RMQ Source

### LINKED ISSUE #455: Event source should not require declaring the exchange and the queue
Right now the event source will [declare an exchange, a queue, and a binding](https://github.com/knative-sandbox/eventing-rabbitmq/blob/95462cfe87b8372512b8f77573e0f47078258477/pkg/adapter/adapter.go#L192-L247). It's possible to specify an existing exchange and queue for the source to use, but it will still redeclare the exchange, queue, and binding.

We should make it possible to skip declaring these resources, since a user may not have permissions to redeclare them, or it may not be possible to redeclare them if they are using options that we don't support.

rikatz:
Signed-off-by: Ricardo Katz <rkatz@vmware.com>

# Changes

This PR adds predeclared behavior in RMQ Source. This way, a user that only got permission to consume a specific queue still can use RabbitMQ Source, declaring the queue to be consumed.

Adapter will try to bind to that queue and get information of it, instead of going through the creation process of a new exchange, queue and binding that exchange to the queue.

/kind enhancement

Fixes #455

**Release Note**

```release-note
RabbitMQ source now can use predeclared queues instead of creating new ones
```

**Docs**

```docs

```

### DIFF
--- a/pkg/adapter/adapter.go
+++ b/pkg/adapter/adapter.go
@@ -72,6 +72,7 @@ type adapterConfig struct {
 	User           string `envconfig:"RABBITMQ_USER" required:"false"`
 	Password       string `envconfig:"RABBITMQ_PASSWORD" required:"false"`
 	Vhost          string `envconfig:"RABBITMQ_VHOST" required:"false"`
+	Predeclared    bool   `envconfig:"RABBITMQ_PREDECLARED" required:"false"`
 	ChannelConfig  ChannelConfig
 	ExchangeConfig ExchangeConfig
 	QueueConfig    QueueConfig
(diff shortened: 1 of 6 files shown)


pkg/adapter/adapter.go
```diff
@@ -72,6 +72,7 @@ type adapterConfig struct {
 	User           string `envconfig:"RABBITMQ_USER" required:"false"`
 	Password       string `envconfig:"RABBITMQ_PASSWORD" required:"false"`
 	Vhost          string `envconfig:"RABBITMQ_VHOST" required:"false"`
+	Predeclared    bool   `envconfig:RABBITMQ_PREDECLARED" required:"false"`
```
benmoss, line 75:
```suggestion
	Predeclared    bool   `envconfig:"RABBITMQ_PREDECLARED" required:"false"`
```

Looks like we don't have any tests that cover Predeclared being true

benmoss (changes requested):

pkg/adapter/adapter.go
rikatz: Thanks, i have added some tests there covering the whole cycle:
* Try to use it, fail (check the error)
* Declare a Queue in fakeServer
* Try again, it should work now

rikatz:
@benmoss I have squashed commits here and probably need authorization to trigger the tests again :)

Thanks!

benmoss:
/approve
/lgtm

thanks @rikatz !

PR #498, opened 2021-11-09
Sink URI resolver tests update

tzununbekov:
Sink resolver updated behavior introduced in knative/pkg#2322 requires `v1.Service` Informer injection to retrieve destination port details.

`v1.Service` informers manually injected and prepopulated with objects if they are required in the tests.

### DIFF
--- a/pkg/reconciler/testing/factory.go
+++ b/pkg/reconciler/testing/factory.go
@@ -23,6 +23,7 @@ import (
 
 	corev1 "k8s.io/api/core/v1"
 	"knative.dev/pkg/configmap"
+	"knative.dev/pkg/injection"
 	"knative.dev/pkg/logging"
 
 	"k8s.io/apimachinery/pkg/api/meta"
@@ -30,6 +31,7 @@ import (
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/types"
 
+	"k8s.io/client-go/rest"
 	"k8s.io/client-go/tools/record"
 
 	"go.uber.org/zap"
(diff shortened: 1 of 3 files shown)


pkg/reconciler/broker/broker_test.go
```diff
@@ -50,6 +52,7 @@ import (
 	v1a1addr "knative.dev/pkg/client/injection/ducks/duck/v1alpha1/addressable"
 	v1b1addr "knative.dev/pkg/client/injection/ducks/duck/v1beta1/addressable"
 	fakekubeclient "knative.dev/pkg/client/injection/kube/client/fake"
+	"knative.dev/pkg/client/injection/kube/informers/core/v1/service"
```
antoineco, line 55: This could be aliased to `serviceinformers` for consistency (and because `service.Key{}` feels kind of off to me 🙂)

antoineco:
@tzununbekov after thinking longer about it, I'm not sure I understand what we need all this.
Wouldn't it be possible to inject a dummy informer using a magic injection import instead? (what you tried few weeks ago at first).
In your previous test, it didn't work because you actually need to read from the informer, but from what I see these tests don't.

tzununbekov:
@antoineco I think that informer injection magic does not work if there is fake client injection involved. `injection.Fake.SetupInformers()`'s [client injection](https://github.com/knative/pkg/blob/release-1.0/injection/informers.go#L101-L105) either overrides ctx values injected by imports of [these](https://github.com/knative-sandbox/eventing-rabbitmq/blob/release-1.0/pkg/reconciler/triggerstandalone/trigger_test.go#L718-L720) clients or I completely misread it. Nevertheless, informer injection magic breaks the tests because fake clients stop returning objects defined in the test cases. If there is a way to make tests work without these dummy informers I'd eagerly use it.

antoineco:
I understand that was the case in `knative/pkg`, but I don't see the Service informer being involved in any tests inside the RabbitMQ repo, did I look in the wrong place?

If I'm not mistaken, the sink resolver only needs to obtain **some** Service informer (just it doesn't panic on `informer.Get(ctx)`), and as long as you don't _use_ that Service informer in your tests, it doesn't matter if the backing ClientSet is empty. Did I get that part wrong? Or maybe your're expecting to add tests that read from the informer in this repo?

tzununbekov:
@antoineco you're correct, Service informer is not used in the tests, we just need to have the context that would not cause a panic in `service.Get(ctx).Lister()`. What currently in the PR is all that I managed to produce to achieve that 😅

benmoss:
https://gist.github.com/benmoss/f2294a47551e08d6674222a9b7c17f86 this seems to fix all the tests more simply

antoineco:
@benmoss ah, good I wasn't dreaming!

tzununbekov:
@benmoss whoa! that worked, thanks!

PR #492, opened 2021-11-05
Use RabbitmqCluster namespace in RabbitmqClusterReference

gvmw:
# Changes

- :bug: &nbsp; Use the RabbitmqCluster namespace in cluster references, otherwise resources will fail to reconcile when RabbitMQ is running in a different namespace

When declaring a new `Broker` of class `RabbitMQBroker`, use the `RabbitmqCluster` namespace in `RabbitmqClusterReference`, otherwise the dependent objects which get declared - i.e. `Exchange`, `Queue`, `Binding` - and are expected to be reconciled by the https://github.com/rabbitmq/messaging-topology-operator will fail to be created.

This has more details: https://github.com/knative-sandbox/eventing-rabbitmq/issues/454

/kind bug

Fixes #454

PR #522, opened 2021-11-19
Source concurrent dispatch

### LINKED ISSUE #508: Event source should support concurrent dispatch
Right now just like https://github.com/knative-sandbox/eventing-rabbitmq/issues/329 the event source only reads one message off the queue at a time and then completes the whole HTTP roundtrip before moving on to the next message. We should support more than one message in-flight at a time.

gabo1208:
# Changes

- 🎁 Concurrent message processing on the source (msg translation to CloudEvent and Dispatching)
- 🧹 Cleaned up concurrency testing for the Adapter and the Dispatcher
- 🧹 Moved the `converToCloudEvent` function to its `message.go` corresponding file

/kind performance

Fixes #508

**Release Note**

```release-note
Now the source Adapter process messages concurrently depending on the channel_config.prefetch_count argument (default to 10)
```


pkg/adapter/adapter.go
```diff
@@ -52,8 +51,8 @@ type ExchangeConfig struct {
 }
 
 type ChannelConfig struct {
-	PrefetchCount int
-	GlobalQos     bool
+	PrefetchCount int  `envconfig:"RABBITMQ_CHANNEL_CONFIG_PREFETCH_COUNT" default:"10" required:"false"`
```
gvmw, line 54: This environment name feels too verbose and implementation-specific. I think that `RABBITMQ_PREFETCH_COUNT` is sufficiently descriptive - wdyt?

Since `PrefetchCount` and `GlobalQos` are both used in the context of `Qos`, I think that it would make sense to rename them to `QosPrefetchCount` and `QosGlobal`.

pkg/adapter/adapter.go
```diff
@@ -281,31 +282,51 @@ func (a *Adapter) PollForMessages(channel *wabbit.Channel,
 
 	msgs, _ := a.ConsumeMessages(channel, queue, logger)
 
+	wg := &sync.WaitGroup{}
+	workerCount := a.config.ChannelConfig.PrefetchCount
```
gvmw, line 286: Prefetch counts of 100 or even 1000 are common recommendations for high-throughput scenarios. Are we sure that we want to start that many goroutines?

Is there going to be a single source per go process? If we are likely to have many sources per go process, this may cause concurrency problems - locking & context switching seem too high to me.

source/README.adoc
```diff
@@ -88,6 +88,8 @@ Sources are Kubernetes objects. In addition to the standard Kubernetes
 | `queue_config.delete_when_unused` {optional} | Boolean
 | `queue_config.exclusive` {optional} | Boolean
 | `queue_config.nowait` {optional} | Boolean
+| `channel_config.prefetch_count` {optional} | Int that limits the https://www.rabbitmq.com/consumer-prefetch.html[Consumer Prefetch Value].
```
gvmw, line 91: This explains the environment variable name chosen here https://github.com/knative-sandbox/eventing-rabbitmq/pull/522/files#r754456187

I am still not convinced that exposing this much RabbitMQ-specific detail is a good idea. If these broker types are meant to be easily interchangeable, then the current implementation-specific approach is going against that principle. How are you thinking about this @benmoss @salaboy ?

pkg/adapter/adapter.go
```diff
@@ -52,8 +51,8 @@ type ExchangeConfig struct {
 }
 
 type ChannelConfig struct {
-	PrefetchCount int
-	GlobalQos     bool
+	PrefetchCount int  `envconfig:"RABBITMQ_CHANNEL_CONFIG_PREFETCH_COUNT" default:"10" required:"false"`
+	GlobalQos     bool `envconfig:"RABBITMQ_CHANNEL_CONFIG_QOS_GLOBAL" required:"false"`
```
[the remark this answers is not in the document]
gabo1208: Since they are specific for the rabbitmq configuration, i think the app prefix is fine here

pkg/adapter/adapter.go
gabo1208: same as in the globalQos

gvmw:
Taking this out for a QA.

gvmw:
I had to jump through the following hoops before I could understand the improvement that this PR brings:
1. `mv config/source/{config-observability,400-config-observability}.yaml` to fix `ko apply -f config/source`
2. add a new example to `samples/source` that is ready to go - [this example](https://github.com/benmoss/knative-scripts/blob/08cc5939a5f465abda7c023696721bf21bc25ec5/rabbit-source.yaml#L14-L48) from @benmoss came in handy
3. revert the source change to see the improvement

It would have been easier for me if this PR already included 1. & 2. I think that building out the samples in this repository can only be a good thing, since anyone can try out various features by following the respective READMEs and running a few `ko` commands. This reminds me that I could have done a better job myself in #492 by adding a README to the multi-namespaces example.

gvmw:
There are a few behaviours that I was not expecting:
- Setting the prefetch to `0` resulted in a prefetch of `10`
- - Setting a prefetch of `-1` was also allowed and resulted in it being set to `10`
- Setting a prefetch of `1000000000` was allowed, and the container was OOMKilled

I would expect the admission webhook to disallow values below `1` and above `10000`. The implications are that RabbitMQ will not allow an infinite number of unacknowledged messages to be sent to this client ([that's what `0` means](https://www.rabbitmq.com/consumer-prefetch.html)), and stop sending more messages if `10000` have not been acknowledged.

gvmw:
We have talked to @gabo1208 about this and concluded that:
- The admission webhook will disallow any prefetch values below `1` and above `10000`
- If prefetch is set to any value between `1` to `10000` the actual value will be respected and the goroutines will be set to the prefetch value too
- Prefetch value defaults to `10` and goroutines default to 10 too.

The reason why we do not allow a prefetch of `0` is so that this client will not be able to receive an infinite number of unacknowledged messages. We don't want to allow users to shoot themselves in the foot.

WDYT @michaelklishin @lukebakken @MarcialRosales @embano1?

lukebakken:
@gvmw it's just a "feeling" but `10000` feels high. But, if that value has been tested and works A-OK, then :shipit:

gvmw:
> @embano1 A bit on the fence here in terms of breaking change and delivery guarantees. Before that change, we basically had basic FIFO semantics (at the expense of low throughput). Not sure, but IMHO we still don't have extensive docs around the delivery behavior of the whole chain (ingress/broker/dispatcher) so this could easily drip off people after the fact, especially when doing upgrades.

That is a great point, let's default prefetch to `1` and goroutines to `1` as well. We should definitely log this when the source starts because the user did not choose a value and communicating defaults is valuable (PoLA - Principle of Least Astonishment).

> @lukebakken I'm glad that you brought it up because `10000` felt high to me too. What would you say would a more reasonable maximum be?

gvmw:
While testing this, we (+@gabo1208) have noticed that the queue does not get deleted when the source is deleted. We are setting the `delete_when_unused: true`, we are investigating why this happens.

lukebakken:
Aha I remembered that we even have a blog article stating there's not much benefit to a prefetch > 1000, at least with quorum queues.

https://blog.rabbitmq.com/posts/2020/05/quorum-queues-and-flow-control-single-queue-benchmarks

gvmw:
This is almost there @gabo1208, apart from the missing `RabbitmqCluster`. This ensures that the sample is self-contained, and `kubectl apply -f samples/source/concurrent-dispatch/` is enough to reproduce everything:

```diff
diff --git a/samples/source/concurrent-dispatch/200-rabbitmq.yaml b/samples/source/concurrent-dispatch/200-rabbitmq.yaml
new file mode 100644
index 00000000..eb6412d7
--- /dev/null
+++ b/samples/source/concurrent-dispatch/200-rabbitmq.yaml
@@ -0,0 +1,8 @@
+---
+apiVersion: rabbitmq.com/v1beta1
+kind: RabbitmqCluster
+metadata:
+  name: rabbitmq
+  namespace: concurrent-dispatch
+spec:
+  replicas: 1
```

gabo1208:
> This is almost there @gabo1208, the only thing missing is `RabbitmqCluster`. This ensures that the sample is self-contained, and `kubectl apply -f samples/source/concurrent-dispatch/` is enough to reproduce everything.
> 
> ```diff
> diff --git a/samples/source/concurrent-dispatch/200-rabbitmq.yaml b/samples/source/concurrent-dispatch/200-rabbitmq.yaml
> new file mode 100644
> index 00000000..eb6412d7
> --- /dev/null
> +++ b/samples/source/concurrent-dispatch/200-rabbitmq.yaml
> @@ -0,0 +1,8 @@
> +---
> +apiVersion: rabbitmq.com/v1beta1
> +kind: RabbitmqCluster
> +metadata:
> +  name: rabbitmq
> +  namespace: concurrent-dispatch
> +spec:
> +  replicas: 1
> ```

The rabbitmq cluster is been created in the same file as the perf test :)

gvmw:
This is the implementation that we have settled on:
- prefetch defaults to `1` - [maintains existing delivery ordering guarantees](https://github.com/knative-sandbox/eventing-rabbitmq/pull/522#issuecomment-981799474)
- when prefetch is set to `0` it will default to `1` - infinite unacknowledged messages are not allowed
- prefetch values above `1000` are not allowed and will fail validation - [thanks @lukebakken](https://github.com/knative-sandbox/eventing-rabbitmq/pull/522#issuecomment-981852719)!

Do you see any issues with any of the above @embano1 @benmoss @xtreme-sameer-vohra?

gvmw:
@gabo1208 I have just noticed that the current version allows for prefetch to be changed post-deploy, but the changes have no effect. Definitely a new bug since yesterday. I am suspecting the new default value.

gabo1208:
> @gabo1208 I have just noticed that the current version allows for prefetch to be changed post-deploy, but the changes have no effect. Definitely a new bug since yesterday. I am suspecting the new default value.

the errors.Also were not working as expected, fixing this

benmoss:
> * when prefetch is set to `0` it will default to `1` - infinite unacknowledged messages are not allowed

@gvmw and I made this patch earlier today to allow us to differentiate the 0 / nil cases: https://gist.github.com/benmoss/0f76aba77912c1df6f8952f091139de3

MarcialRosales:
> This is the implementation that we have settled on:
> 
> * prefetch defaults to `1` - [maintains existing delivery ordering guarantees](https://github.com/knative-sandbox/eventing-rabbitmq/pull/522#issuecomment-981799474)
> * when prefetch is set to `0` it will default to `1` - infinite unacknowledged messages are not allowed
> * prefetch values above `1000` are not allowed and will fail validation - [thanks @lukebakken](https://github.com/knative-sandbox/eventing-rabbitmq/pull/522#issuecomment-981852719)!

@gvmw  First of all, sorry for the late reply. Totally agree with the settlement. At first, I was a bit hesitant of putting an upper limit of 1000. However based on the blog post that @lukebakken mentioned maybe having that limit makes sense. Also in all the PerfTest simulations that i did for customers I don't recall exceeding 500. Beyond that point we did not see any improvements. We did notice improvements when we acknowledged in batches, but that is a separate story.

gvmw:
Great feedback @MarcialRosales, thank you!

samples/source/concurrent-dispatch/200-perf-test.yaml
```diff
@@ -0,0 +1,78 @@
+# RabbitMQ cluster used by the Source
+apiVersion: rabbitmq.com/v1beta1
+kind: RabbitmqCluster
+metadata:
+  name: rabbitmq
+  namespace: concurrent-dispatch
+spec:
+  replicas: 1
+---
```
gvmw, line 9: This should be split into a different YAML file. `200-rabbitmq.yaml` seems appropriate.

pkg/adapter/adapter.go
embano1: What do you mean by "sources"? In general, Goroutines are lightweight and having thousands of them does not necessarily impact (thread) context switching. Not sure where we'd have locking bottlenecks in this code path.

gvmw:
Taking this for another spin on GKE - hopefully the last one!

gvmw:
If the Rabbitmq that the source is pointing to is not available, the source will crash, which I think is OK because the error message is descriptive even though it looks scary:

image

image

As soon as RabbitMQ is running and the source pod gets re-scheduled, everything looks good again:
image

## Sanity check:

- the channel used by the source channel defaults to `1`, so the behaviour does not change @embano1.
- prefetch can be set to any value between `1` and `1000` (any other values are not allowed by the admission webhook)
- modifying the prefetch is not allowed, it's an immutable field (same behaviour as before)

This last point could be improved. When the prefetch changes, the source client should reconnect (re-scheduling the pod is enough) so that the new value is used in the channel that is created on connecting. With the current implementation, the source needs to be deleted and re-declared with the new value. This workaround is OK for now, since the queue backing the source does not need to be deleted.

/lgtm
/approve

embano1:
> If the Rabbitmq that the source is pointing to is not available, the source will crash, which I think is OK

ACK

> the channel used by the source channel defaults to `1` [...] When the prefetch changes, the source client should reconnect

Not sure I follow what `source` is meant here. Isn't prefetch relevant to the worker/dispatcher (`trigger`) only?

gvmw:
> missed that this PR is for the RabbitMQ `source`

Here is the `Trigger` / `Dispatcher` equivalent https://github.com/knative-sandbox/eventing-rabbitmq/pull/536

PR #525, opened 2021-11-22
Setup local host & run end-to-end tests with a single command

gvmw:
# Changes

- [x] Add `Makefile` to make discovery and running of dev-related tasks easy and straightforward
- [x] Remove script that is used for e2e setup (this differs from the CI workflow)
- [x] Use make targets in `.github/workflows/kind-e2e.yaml`
- [x] Use make targets in `.github/workflows/kind-conformance-standalone.yaml`
- [x] Use make targets in `.github/workflows/kind-conformance.yaml`

I had to figure out how to setup my local environment for running end-to-end tests, and while there is a script in this repo that does the same thing, it is outdated, incomplete and difficult to find. While all these are solvable problems, the discovery and accessibility of portions of this script are more difficult to tackle. Multiple individual scripts that are wrapped in a single one with shared functions is a common way of solving this, but it takes very little for this to become hard to maintain and own by newcomers. Having used both approaches in multiple projects over the last decade, this is a better overall experience.

## How do I use this?

Given a clean Linux host with `make`, Golang & Docker Engine installed, this is how it works:

```sh
make

If this is your first time running this, remember to run: make .env && source .env
Now just type make <TAB> to enjoy shell autocompletion
By the way, m is an alias for make

Here is a list of all the make targets that you can run, e.g. make test-e2e or m test-e2e

...
```

Let's focus on the first time setup:

```sh
make .env
make --file Makefile --no-print-directory env SILENT="1>/dev/null 2>&1" > .env
. .env
```

To check if this is setup correctly, let's check `make` autocompletion (assumes your shell is bash):

```sh
make <TAB>
bash-autocomplete                   install                             kind-cluster                        releases-kind                       test-e2e-publish
curl                                install-cert-manager                kn                                  releases-kn                         test-e2e-source
env                                 install-knative-eventing            ko                                  releases-ko                         test-unit
envsubst                            install-knative-serving             kubectl                             releases-kubectl                    test-unit-uncached
gcloud                              install-rabbitmq-cluster-operator   releases-envsubst                   test-compilation
gh                                  install-rabbitmq-topology-operator  releases-gcloud                     test-conformance
go-dep-update                       k9s                                 releases-gh                         test-e2e
help                                kind                                releases-k9s                        test-e2e-broker
```

Great! Let's install all the things: `kubectl`, `kind` and then KiND `eventing-rabbitmq-e2e` cluster with everything that is needed to run tests:

```sh
# 💡 If your KUBECONFIG is not pointing to a running K8s cluster, run:
# make kubeconfig
make install
...
deployment.apps/rabbitmq-webhook created
service/rabbitmq-webhook created
configmap/config-logging created
configmap/config-observability created
/root/github.com/gvmw/eventing-rabbitmq/bin/kubectl-1.20.7-linux-amd64 wait --for=condition=available deploy/pingsource-mt-adapter --timeout=60s --namespace knative-eventing
deployment.apps/pingsource-mt-adapter condition met
```

> ASIDE: In case you get any failures related to `docker-credential-gcloud`, delete the `~/.docker/config.json` file

## How do I run end-to-end tests?

To run end-to-end tests, which is what I had to do in the context of https://github.com/knative-sandbox/eventing-rabbitmq/pull/492 (this PR is based on that one), I now only need to run the following:

```sh
make test-e2e

# I can run just a subset of tests with e.g.:
make test-e2e-broker
```

## How do I run conformance tests?

👉🏻 @gvmw to do this next

## What was in scope for this PR?

- only support Linux & Darwin systems (does not support Windows)
- only supports x84_64 (does not support latest Macs with ARM silicon)
- defaults to the minimum supported K8s version (adding support for multiple K8s versions was out of scope)

This was only tested on Linux Ubuntu 20.04.3 LTS, and should be tested on Darwin on Intel (I don't run Docker on Darwin). Supporting Darwin on ARM should not be difficult, but as I don't have access to one, I cannot test it.

---

By the way, there is a related Knative Proposal regarding this topic: [Google Doc: Proposal Makefile for Knative](https://docs.google.com/document/d/1GGHxXZ4mEKUa5CtNvheK4MBqOlgr2U1kL4e2gVxHOVc/edit#heading=h.9pt7qa7d30fo)

cc @benmoss @nader-ziada @n3wscott @salaboy


n3wscott (commented):
Makefiles do not work out of the box on windows and in Knative, we have never wanted to add makefiles to the project. Cc @evankanderson

gabo1208 (changes requested):

Makefile
```diff
@@ -0,0 +1,323 @@
...
+.PHONY: bash-autocomplete
+bash-autocomplete:
+	@echo "$(BASH_AUTOCOMPLETE)"
+env:: bash-autocomplete
+
+.PHONY: dep-update
+dep-update: ## Update any dependency
+	@printf "Update dep in go.mod by running e.g. $(BOLD)go get -d github.com/rabbitmq/messaging-topology-operator@v1.2.1$(RESET)\n" \
+	; read -rp " (press any key when done)" -n 1
+	$(CURDIR)/hack/update-deps.sh
```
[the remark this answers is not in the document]
gvmw: There were situations when I had to run `update-deps` before `update-codegen`, so this accounts for that.

Maybe running `update-deps` before & after `update-codegen` is the way to go?

gvmw:
thanks for the feedback @n3wscott !

> Makefiles do not work out of the box on windows and in Knative

If supporting Windows as a development platform is important, then something like GitHub Codespaces or Gitpod is the way to achieve that in my opinion. As an alternative, I think a more modern implementation of make like [just](https://github.com/casey/just) may work too.

I am not sure how to address the "in Knative" part.

> we have never wanted to add makefiles to the project

If someone can show me a better alternative to `make e2e-setup` and `make test-e2e`, I would really like to see it. I am especially interested in a solution that is self-documenting and is continuously validating, meaning that we can run it in GitHub Actions.

gvmw:
I had a look at https://github.com/benmoss/knative-scripts and I think that all of it should live in this repository. I also think that bringing it up in the context of this PR is relevant since we could expose all those scripts as self-descriptive make targets. This is what that would look like in practice:

```
build                               Build binaries with e2e tags
dep-update                          Update any dependency
e2e-setup                           Setup environment for e2e testing
env                                 Configure shell env - eval "$(make env)" OR rm .env && make .env && source .env
install-cert-manager                Install Cert Manager - dependency of RabbitMQ Topology Operator
install-knative-eventing            Install Knative Eventing
install-knative-eventing-rabbitmq   Install dev Knative Eventing RabbitMQ - also installs Knative Eventing
install-rabbitmq-cluster-operator   Install RabbitMQ Cluster Operator
install-rabbitmq-topology-operator  Install RabbitMQ Topology Operator
k9s                                 Terminal ncurses UI for K8s
test-e2e-broker                     Run Broker end-to-end tests
test-e2e-publish                    Run TestKoPublish end-to-end tests
test-e2e-source                     Run Source end-to-end tests
test-unit                           Run unit tests
```

Makefile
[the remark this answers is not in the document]
benmoss: Yeah, if you're updating one of the RabbitMQ deps you'll need to run `update-deps` to update the dependency itself, and then `update-codegen` to update the generated clients, etc. We run `update-deps` at the end of codegen because in theory the codegen library might change and introduce new dependencies _in the generated code itself_.

This makes it hard to automate in a one-liner without always doing the most amount of redundant work possible. We could in theory have a dependency graph of this but that sounds hard.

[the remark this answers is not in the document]
gvmw: Updating deps twice seems like the simplest approach, even if it's not the fastest one. Your comment makes me realise that this still stands.

gvmw:
Now that #492 is merged, this is a focused changed.

I have been using this for about 2 weeks and it works well. There is a single thing missing - a paragraph added to README - but otherwise this is ready to be merged. Try it out locally and let me know how well it works for you in practice. If this change is a net positive for the people that work with this repository every day, my proposal would be to merge it. If it doesn't work for you, capture your reasons - a sentence is OK - and let's move on.

Makefile
```diff
@@ -0,0 +1,363 @@
...
+export KIND_K8s_DIGEST
+
+.PHONY: kind-cluster
+kind-cluster: | $(KIND) $(ENVSUBST)
+	( $(KIND) get clusters | grep $(KIND_CLUSTER_NAME) ) \
+	|| ( cat $(CURDIR)/test/e2e/kind.yaml \
+	     | $(ENVSUBST_SAFE) \
+	     | $(KIND) create cluster --name $(KIND_CLUSTER_NAME) --config - )
+
+$(KUBECONFIG): | kind-cluster $(KUBECONFIG_DIR)
```
benmoss, line 288: I don't think we should make everything dependent on `kind-cluster`, it means they aren't reusable against an already-existing cluster (GKE, a differently named kind cluster, etc). They should just assume a valid kubernetes context has been set.

Makefile
```diff
@@ -0,0 +1,363 @@
...
+	tar zxf $(GH_DIR).tar.gz -C $(LOCAL_BIN)
+	touch $(GH)
+	chmod +x $(GH)
+	$(GH) version | grep $(GH_VERSION)
+	ln -sf $(GH) $(LOCAL_BIN)/gh
+.PHONY: gh
+gh: $(GH)
+.PHONY: releases-gh
+releases-gh:
+	$(OPEN) $(GH_RELEASES)
```
benmoss, line 184: This doesn't seem to work for me
```
❯ make releases-gh
https://github.com/cli/cli/releases
bash: https://github.com/cli/cli/releases: No such file or directory
make: *** [Makefile:184: releases-gh] Error 127
```

Makefile
```diff
@@ -0,0 +1,363 @@
...
+	@printf "$(WARN)Remember to run $(BOLD)make e2e-setup$(RESET)$(WARN) if needed$(RESET)\n"
+	go test -v -race -count=1 -timeout=15m -tags=e2e ./test/e2e/... -run 'Test.*Source.*' \
+	| grep -v "no test files"
+
+.PHONY: test-e2e
+test-e2e: test-e2e-publish test-e2e-broker test-e2e-source
+
+BUILD_TAGS=e2e
+.PHONY: build
+build: ## Build binaries with e2e tags
```
benmoss, line 360: I wouldn't call this `build`, since it doesn't actually build any binaries. It's more a test that things *can* build, the compiler makes no warnings. We use `go test` to test that tests also compile, regular `go build` doesn't build tests apparently.

Makefile
```diff
@@ -0,0 +1,363 @@
...
+.PHONY: test-e2e-broker
+test-e2e-broker: ## Run Broker end-to-end tests
+	@printf "$(WARN)Remember to run $(BOLD)make e2e-setup$(RESET)$(WARN) if needed$(RESET)\n"
+	@printf "$(WARN)$(BOLD)rabbitmqcluster$(RESET)$(WARN) has large resource requirements 🤔$(RESET)\n"
+	go test -v -race -count=1 -timeout=15m -tags=e2e ./test/e2e/... -run 'Test.*Broker.*' \
+	| grep -v "no test files"
+
+.PHONY: test-e2e-source
+test-e2e-source: ## Run Source end-to-end tests
+	@printf "$(WARN)Remember to run $(BOLD)make e2e-setup$(RESET)$(WARN) if needed$(RESET)\n"
```
benmoss, line 351: Telling someone to do something "if needed" is kind of roundabout, how do I know if it's needed?

Makefile
```diff
@@ -0,0 +1,363 @@
...
+KNATIVE_EVENTING_VERSION ?= 1.0.0
+install-knative-eventing: | $(KUBECONFIG) $(KUBECTL) ## Install Knative Eventing
+	$(KUBECTL) $(K_CMD) --filename \
+		https://github.com/knative/eventing/releases/download/knative-v$(KNATIVE_EVENTING_VERSION)/eventing-crds.yaml
+	$(KUBECTL) $(K_CMD) --filename \
+		https://github.com/knative/eventing/releases/download/knative-v$(KNATIVE_EVENTING_VERSION)/eventing-core.yaml
+	$(KUBECTL) wait --for=condition=available deploy/eventing-controller --timeout=30s --namespace $(SYSTEM_NAMESPACE)
+	$(KUBECTL) wait --for=condition=available deploy/eventing-webhook --timeout=30s --namespace $(SYSTEM_NAMESPACE)
+
+install-knative-eventing-rabbitmq: | $(KUBECONFIG) $(KO) install-knative-eventing ## Install dev Knative Eventing RabbitMQ - also installs Knative Eventing
```
benmoss, line 320: Maybe just `install` for this one, since this is the for this repo's code?

Makefile
benmoss: I don't actually know what this part of the Makefile is useful for, I don't see `$(OPEN)` defined anywhere

gvmw: https://github.com/knative-sandbox/eventing-rabbitmq/pull/525/commits/8870f7360f1c300d31d68c0f6c0e70a1eeeb864a fixes it, thanks!

Makefile
gvmw: You are right, this makes more sense: https://github.com/knative-sandbox/eventing-rabbitmq/pull/525/commits/b05b259667e92edb460060e155d0651bcc30f428

How do you think that we should hint to the fact `KUBECONFIG` can be set in the environment, and then no kind cluster will be created?

Makefile
gvmw: This is the `Build` step from this GitHub Action workflow: https://github.com/knative-sandbox/eventing-rabbitmq/blob/23a39265b91a2e9f1b4729f12a6f6f94a5ac906f/.github/workflows/knative-go-build.yaml#L51-L65

What do you think that we should call it instead?

Makefile
gvmw: Fixed in https://github.com/knative-sandbox/eventing-rabbitmq/pull/525/commits/0a4972f9b9330b190643cb50df1b2ac575a8aacc

Makefile
benmoss: Maybe test-compilation?

Makefile
```diff
@@ -0,0 +1,378 @@
...
+OPEN := open
+else
+OPEN := xdg-open
+endif
+
+GCLOUD_SDK_VERSION := 365.0.0
+GCLOUD_BIN := gcloud-$(GCLOUD_SDK_VERSION)-$(PLATFORM)-x86_64
+GCLOUD := $(LOCAL_BIN)/$(GCLOUD_BIN)
+GCLOUD_SDK_FILE := google-cloud-sdk-$(GCLOUD_SDK_VERSION)-$(PLATFORM)-x86_64.tar.gz
+GCLOUD_SDK_URL := https://dl.google.com/dl/cloudsdk/channels/rapid/downloads/$(GCLOUD_SDK_FILE)
```
cardil, line 54: I don't like downloading tools directly from the network. I have them installed on my machine, securely. Without checking for consistency (signs, checksums) you are introducing a vulnerability into Knative supply chain.

Makefile
```diff
@@ -0,0 +1,378 @@
...
+# https://github.com/knative/eventing/releases
+install-knative-eventing: | $(KUBECONFIG) $(KUBECTL) ## Install Knative Eventing
+	$(KUBECTL) $(K_CMD) --filename \
+		https://github.com/knative/eventing/releases/download/knative-v$(KNATIVE_VERSION)/eventing-crds.yaml
+	$(KUBECTL) $(K_CMD) --filename \
+		https://github.com/knative/eventing/releases/download/knative-v$(KNATIVE_VERSION)/eventing-core.yaml
+	$(KUBECTL) wait --for=condition=available deploy/eventing-controller --timeout=60s --namespace $(EVENTING_NAMESPACE)
+	$(KUBECTL) wait --for=condition=available deploy/eventing-webhook --timeout=60s --namespace $(EVENTING_NAMESPACE)
+
+install: | $(KUBECONFIG) $(KO) install-knative-serving install-knative-eventing install-rabbitmq-cluster-operator install-rabbitmq-topology-operator ## Install local dev Knative Eventing RabbitMQ - manages all dependencies, including K8s components
```
cardil, line 347: This script is sooo long. Why will be maintaining this? There are 20-30 knative repos. Are we realy like to copy and paste this so many times, and change couple of things, that needs adjusting, in between?!?

gvmw:
> -1 from me. Effectively merging couple bash scripts into one Makefile.

Thank you for the feedback!

If you mean shell commands, then I agree. There are no function calls, includes, or libraries, it's just a bunch of commands that run selectively & sequentially.

> If we adopt makefile we should compose them from reusable parts to keep it manageable.

I would keep this as simple as it can be, and see how well it works in practice for everyone on the team, across a few weeks, before investing more in this direction. I think seeing the [Magefile](https://magefile.org/) alternative to this would be super valuable, so if you want to contribute that version, I would be all for it. I think there is a lot of benefit in that approach, but I don't have your context, and being able to compare both approaches side-by-side would really help me fill in the missing gaps.

Makefile
gvmw: I definitely think that this area should be improved.

I also think the zero-trust approach promoted by chainguard.dev is the right way to go, and considering the people that are behind this idea - @mattmoor @vaikas @n3wscott - I think that they would agree with this too.

Considering that we adopt the same approach in our existing GitHub Actions, and we suggest the same in the [Knative development guide](https://github.com/knative/eventing/blob/main/DEVELOPMENT.md), I don't see the approach in this PR being worse. Having said that, I'm glad that this is on your radar and I also think that we should close this gap as soon as possible, but not part of this PR.

Makefile
gvmw: I understand that your perspective is the wider Knative ecosystem and I must agree that I have not thought about the big picture since my focus was to get this working in the context of this repository. I recognise the tension between _Make it work_ vs _Make it right_ and while I welcome the second challenge, the focus is on small scope and trying to understand if this approach will even work.

I must emphasise that this is not a script, it's a collection of commands chained together. The distinction may seem insignificant, but it's an important mind shift for approaching Make the right way. This example illustrates Make approached from a different perspective which is valid for that context, but would be the wrong approach for Knative in my opinion:

https://github.com/ninenines/erlang.mk/blob/d80984c1036ea81eb1f44b8d7cde85fc09b5e3c0/core/deps.mk#L197-L248

Makefile
gvmw: Fixed in https://github.com/knative-sandbox/eventing-rabbitmq/pull/525/commits/e5e6bbf2d50b0e2c8f432022e3429cb2e187091a

Makefile
benmoss: I think we should have kind cluster creation as an explicit step, not automatic/implicit

gvmw: I can see how it would make sense to fail if `KUBECONFIG` is not set, rather than default to creating a KiND cluster. Are you thinking the same thing?

gvmw:
For some reason, brokers and triggers are failing to reconcile correctly when running conformance tests via `make test-conforrmance`:

```
NAMESPACE       NAME              URL   AGE    CONDITIONS   READY   REASON
test-crrvhydv   broker-cvpkdtpm         149m   0 OK / 7     False   ReconcileFailure : using secret is not supported with this controller
test-crrvhydv   broker-kpqkubzt         139m   0 OK / 7     False   ReconcileFailure : using secret is not supported with this controller
test-crrvhydv   default                 154m   0 OK / 7     False   ReconcileFailure : using secret is not supported with this controller
```

Has anyone seen this before?

I suspect it's an environment variable not being propagated correctly, but it's possible that one of the deps changed versions. Investigating...

Makefile
benmoss: No, it's possible to not have `KUBECONFIG` set and still be valid, for example if you have a `~/.kube/config` file.

benmoss:
Looks like you're test-conformance-standalone is not installing the standalone broker

gvmw:
The conformance tests as well as the e2e tests are passing locally, using the same versions that are failing in GitHub Actions. I think that we may hitting against the limit of the free GitHub-hosted runners. It may be that we are installing more things by default, or that we are moving through the stages quicker, but this is what the local runs look like:

```
time make install 2>&1 | tee make.install.log
# Everything works on & completes in 3m 10s
# 👇🏻 find the output attached 👇🏻
```
[make.install.log](https://github.com/knative-sandbox/eventing-rabbitmq/files/7729243/make.install.log)

```
time make test-e2e 2>&1 | tee make.e2e.log
# Everything passes & completes in Nm Ns
# 👇🏻 find the output attached 👇🏻
```

```
time make install-standalone 2>&1 | tee make.install-standalone.log
# Everything works on & completes in Nm Ns
# 👇🏻 find the output attached 👇🏻
```

```
time make test-conformance-standalone 2>&1 | tee make.conformance-standalone.log
# Everything works on & completes in Nm Ns
# 👇🏻 find the output attached 👇🏻
```

Should we keep this addition as dev-only, and solve the GitHub Actions (CI) issues separately? This is unlikely to get merged otherwise.

gvmw:
Now that we are not installing Knative Serving in Conformance Tests, everything is working fine, so it was a load issue.

One last change and then this is ready for Q&A. We need a Darwin Intel host with - can you help @ikvmw ?

PR #533, opened 2021-12-02
Use policies for broker/trigger queue creation

### LINKED ISSUE #448: Updating a trigger's deadLetterSink puts the queue into a bad state
There's not a way to update the dead letter exchange of a queue, so I think if we encounter the following scenario, the only option other than immutability would be to delete and recreate the queue.

- Declare a broker with a dead letter sink
- Declare a trigger with a dead letter sink
- Remove the DLS from the trigger

You'll see that the queue doesn't change to point to the broker's DLX, and if you inspect the queue CRD you'll see something like:

```yaml
status:
  conditions:
  - lastTransitionTime: "2021-10-04T15:59:05Z"
    message: 'Error 400 (bad_request): inequivalent arg ''x-dead-letter-exchange''
      for queue ''t.default.default.a8b1707f-b6aa-42e4-8707-664137715fcc'' in vhost
      ''/'': received ''b.default.default.dlx.51802894-5fff-43a6-949b-15b500b0c60e''
      but current is ''t.default.default.dlx.a8b1707f-b6aa-42e4-8707-664137715fcc'''
    reason: FailedCreateOrUpdate
    status: "False"
    type: Ready
```
### Blocker
- [ ] https://github.com/knative-sandbox/eventing-rabbitmq/issues/264

benmoss:
# Changes

- :bug: Allow users to change the DLS on a trigger, which previously was not possible because queue properties are immutable

/kind bug

Fixes #448

**Release Note**

```release-note
:page_facing_up: Removing the dead letter sink on a trigger will now properly fall back to the broker's dead letter sink, if one is defined
```

### DIFF
--- a/config/broker/200-rabbitmq-broker-clusterrole.yaml
+++ b/config/broker/200-rabbitmq-broker-clusterrole.yaml
@@ -52,6 +52,7 @@ rules:
     - bindings
     - exchanges
     - queues
+    - policies
     verbs:
     - "get"
     - "list"
--- a/pkg/rabbit/service.go
+++ b/pkg/rabbit/service.go
@@ -73,26 +73,47 @@ func (r *Rabbit) ReconcileExchange(ctx context.Context, args *resources.Exchange
 func (r *Rabbit) ReconcileQueue(ctx context.Context, args *triggerresources.QueueArgs) (Result, error) {
 	logging.FromContext(ctx).Info("Reconciling queue")
 
-	queueName := args.Name
 	want := triggerresources.NewQueue(ctx, args)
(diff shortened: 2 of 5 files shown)


ikvmw (changes requested):
If I understand correctly, this PR has two parts - one about queue policy and another is relatively big refactoring. Although I see the point of the refactoring, I think it should be split and submitted as another PR. 

The queue policy part looks ok. Also learnt some cool things like equality/semantic.go

benmoss:
/hold

wait for https://github.com/knative-sandbox/eventing-rabbitmq/pull/552

ikvmw:
LGTM

ikvmw (approved):

gabo1208 (approved):

PR #538, opened 2021-12-06
Fix rekt compile errors

pierDipi:
https://github.com/knative-sandbox/reconciler-test/pull/261
has breaking changes that this patch fixes.

/assign @benmoss @gabo1208


pierDipi:
One more, 
/assign @salaboy

PR #543, opened 2021-12-08
WIP: Add publisher confirms to ingress 

### LINKED ISSUE #334: Ensure Broker Ingress uses Publisher-Confirm.

ikvmw:
# Changes

-
-
-

/kind <kind>

Fixes #334 

**Release Note**

```release-note

```

**Docs**

```docs

```

### DIFF
--- a/cmd/ingress/channel.go
+++ b/cmd/ingress/channel.go
@@ -1,0 +1,107 @@
+/*
+   Copyright 2021 The Knative Authors
+
+   Licensed under the Apache License, Version 2.0 (the "License");
+   you may not use this file except in compliance with the License.
+   You may obtain a copy of the License at
+
+http://www.apache.org/licenses/LICENSE-2.0
+
+   Unless required by applicable law or agreed to in writing, software
+   distributed under the License is distributed on an "AS IS" BASIS,
+   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+   See the License for the specific language governing permissions and
+   limitations under the License.
+*/
+
+package main
+
+import (
+	"log"
+	"sync"
+
+	amqp "github.com/rabbitmq/amqp091-go"
+)
+
+type Channel struct {
+	channel        *amqp.Channel
+	counter        uint64
+	mutex          sync.Mutex
+	confirmsMap    sync.Map
+	pconfirms_chan chan amqp.Confirmation
+}
+
+func openChannel(conn *amqp.Connection) *Channel {
+	a_channel, err := conn.Channel()
+	if err != nil {
+		log.Fatalf("failed to open a channel: %s", err)
+	}
+
+	channel := &Channel{channel: a_channel, counter: 1, pconfirms_chan: make(chan amqp.Confirmation)}
+
+	go channel.confirmsHandler()
+
+	a_channel.NotifyPublish(channel.pconfirms_chan)
+	// noWait is false
+	if err := a_channel.Confirm(false); err != nil {
+		log.Fatalf("faild to switch connection channel to confirm mode: %s", err)
+	}
+
+	return channel
+}
+
+func (channel *Channel) publish(exchange string, headers amqp.Table, bytes []byte) (chan amqp.Confirmation, error) {
+	channel.mutex.Lock()
+	defer channel.mutex.Unlock()
+
+	cchan := make(chan amqp.Confirmation, 1)
+
+	channel.confirmsMap.Store(channel.counter, cchan)
+
+	if err := channel.channel.Publish(
+		exchange,
+		"",    // routing key
+		false, // mandatory
+		false, // immediate
+		amqp.Publishing{
+			Headers:     headers,
+			ContentType: "application/json",
+			Body:        bytes,
+		}); err != nil {
+		channel.confirmsMap.LoadAndDelete(channel.counter)
+		return nil, err
+	}
+
+	channel.counter++
+
+	return cchan, nil
+}
+
+func (channel *Channel) confirmsHandler() func() error {
+	return func() error {
+		for confirm := range channel.pconfirms_chan {
+			channel.handleConfirm(confirm)
+		}
+		return nil
+	}
+}
+
+func (channel *Channel) handleConfirm(confirm amqp.Confirmation) {
+	// NOTE: current golang driver resequences confirms and sends them one-by-one
+	value, present := channel.confirmsMap.LoadAndDelete(confirm.DeliveryTag)
+
+	if !present {
+		// really die?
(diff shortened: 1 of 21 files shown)


cmd/ingress/main.go
```diff
@@ -134,17 +147,40 @@ func (env *envConfig) send(event *cloudevents.Event) (int, error) {
...
+		return http.StatusAccepted, nil
+	} else {
+		return http.StatusServiceUnavailable, errors.New("message was not confirmed")
+	}
+}
+
+func (env *envConfig) sendMessage(headers amqp.Table, bytes []byte) (chan amqp.Confirmation, error) {
+	crc := atomic.AddUint64(&env.rc, 1)
+
+	channel := env.channels[crc%uint64(env.cc)]
```
benmoss, line 169: We should use a Go channel and have all these Channels select on it rather than trying to do this load balancing logic

cmd/ingress/channel.go
```diff
@@ -0,0 +1,107 @@
...
+		return nil, err
+	}
+
+	channel.counter++
+
+	return cchan, nil
+}
+
+func (channel *Channel) confirmsHandler() func() error {
+	return func() error {
```
benmoss, line 81: I don't think you need to make this an anonymous function, just have this be the signature of `confirmsHandler`

cmd/ingress/channel.go
```diff
@@ -0,0 +1,107 @@
...
+	}
+}
+
+func (channel *Channel) handleConfirm(confirm amqp.Confirmation) {
+	// NOTE: current golang driver resequences confirms and sends them one-by-one
+	value, present := channel.confirmsMap.LoadAndDelete(confirm.DeliveryTag)
+
+	if !present {
+		// really die?
+		log.Fatalf("Received confirm for unknown send. DeliveryTag: %d", confirm.DeliveryTag)
```
benmoss, line 95: I would `Warn` but not `Fatal` here, this shouldn't be possible unless there's a bug in RabbitMQ

benmoss (commented):
I think we need to step back and improve the way the amqp library handles confirms rather than working around it like this. It's pretty hard to reason about this solution with all the levels of concurrency that we're working with here.

I am thinking something like if `channel.Publish` returned a promise-type object that we would be able to push a lot of this complexity into the library and hide it from users.

cmd/ingress/channel.go
```diff
@@ -0,0 +1,107 @@
...
+	pconfirms_chan chan amqp.Confirmation
+}
+
+func openChannel(conn *amqp.Connection) *Channel {
+	a_channel, err := conn.Channel()
+	if err != nil {
+		log.Fatalf("failed to open a channel: %s", err)
+	}
+
+	channel := &Channel{channel: a_channel, counter: 1, pconfirms_chan: make(chan amqp.Confirmation)}
```
embano1, line 40: we use unbuffered channel for confirmation. docs for confirmation channel in NotifyPublish state it should be buffered to accomodate for multiple in-flights, otherwise risking deadlocks. Could we run into this scenario with an unbuffered channel?

cmd/ingress/channel.go
```diff
@@ -0,0 +1,107 @@
...
+
+	return cchan, nil
+}
+
+func (channel *Channel) confirmsHandler() func() error {
+	return func() error {
+		for confirm := range channel.pconfirms_chan {
+			channel.handleConfirm(confirm)
+		}
+		return nil
```
embano1, line 85: a log debug line could not harm here to understand that the chan was closed, e.g. due to underlying RMQ channel closed

cmd/ingress/main.go
```diff
@@ -134,17 +147,40 @@ func (env *envConfig) send(event *cloudevents.Event) (int, error) {
...
+
+	if confirmation.Ack {
+		return http.StatusAccepted, nil
+	} else {
+		return http.StatusServiceUnavailable, errors.New("message was not confirmed")
+	}
+}
+
+func (env *envConfig) sendMessage(headers amqp.Table, bytes []byte) (chan amqp.Confirmation, error) {
+	crc := atomic.AddUint64(&env.rc, 1)
```
embano1, line 167: looks like this means the first message will bump counter to 1 and hence env.channels[0] would never be used? if so, you might change your slice indexing in line 175 accordingly

cmd/ingress/main.go
```diff
@@ -134,17 +147,40 @@ func (env *envConfig) send(event *cloudevents.Event) (int, error) {
...
-		"",    // routing key
-		false, // mandatory
-		false, // immediate
-		amqp.Publishing{
-			Headers:     headers,
-			ContentType: "application/json",
-			Body:        bytes,
-		}); err != nil {
+
+	ch, err := env.sendMessage(headers, bytes)
```
embano1, line 151: nit: confirmCh

cmd/ingress/channel.go
```diff
@@ -0,0 +1,107 @@
...
+	// noWait is false
+	if err := a_channel.Confirm(false); err != nil {
+		log.Fatalf("faild to switch connection channel to confirm mode: %s", err)
+	}
+
+	return channel
+}
+
+func (channel *Channel) publish(exchange string, headers amqp.Table, bytes []byte) (chan amqp.Confirmation, error) {
+	channel.mutex.Lock()
```
embano1, line 54: This basically means that every message published will lock the particular channel, i.e. serialize the logic and constrain parallelism to the number of defined channels, correct?

cmd/ingress/channel.go
```diff
@@ -0,0 +1,107 @@
...
+	}
+
+	channel := &Channel{channel: a_channel, counter: 1, pconfirms_chan: make(chan amqp.Confirmation)}
+
+	go channel.confirmsHandler()
+
+	a_channel.NotifyPublish(channel.pconfirms_chan)
+	// noWait is false
+	if err := a_channel.Confirm(false); err != nil {
+		log.Fatalf("faild to switch connection channel to confirm mode: %s", err)
```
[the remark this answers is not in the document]
embano1: nit: typo faild

embano1 (commented):
Overall I found the card hard to review for correctness due to the concurrency and **non-transactional** locking at play. I also don't see how this code is more performant than the proposed sync.Pool take/return Channel proposed by @bmoss.

Can you please explain why this complexity is justified? Perhaps I'm not understanding the optimizations at play.

PR #552, opened 2021-12-14
Refactor rabbit service to return Results

benmoss:
# Changes

- :broom: Change the contract of the RabbitService interface to return `Result` objects rather than the CRD
- :broom: Remove exchange, binding, and queue informers and just use the rabbitmq typed client to do CRUD operations from within the RabbitService

/kind cleanup

Requested to be split out of #533 by @ikvmw

### DIFF
--- a/pkg/reconciler/broker/broker.go
+++ b/pkg/reconciler/broker/broker.go
@@ -36,10 +36,10 @@ import (
 	"knative.dev/pkg/logging"
 	"knative.dev/pkg/network"
 
+	"knative.dev/eventing-rabbitmq/pkg/rabbit"
 	naming "knative.dev/eventing-rabbitmq/pkg/rabbitmqnaming"
 	"knative.dev/eventing-rabbitmq/pkg/reconciler/broker/resources"
 	triggerresources "knative.dev/eventing-rabbitmq/pkg/reconciler/trigger/resources"
-	"knative.dev/eventing-rabbitmq/third_party/pkg/apis/rabbitmq.com/v1beta1"
 	rabbitclientset "knative.dev/eventing-rabbitmq/third_party/pkg/client/clientset/versioned"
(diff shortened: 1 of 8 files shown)


pkg/reconciler/broker/broker.go
```diff
@@ -277,25 +270,21 @@ func (r *Reconciler) reconcileUsingCRD(ctx context.Context, b *eventingv1.Broker
...
-			MarkExchangeFailed(&b.Status, "ExchangeFailure", fmt.Sprintf("exchange %q is not ready", exchange.Name))
-			return nil
-		}
+	if !exchange.Ready {
+		logging.FromContext(ctx).Warnf("Exchange %q is not ready", exchange.Name)
+		MarkExchangeFailed(&b.Status, "ExchangeFailure", fmt.Sprintf("exchange %q is not ready", exchange.Name))
+		return nil
 	}
 	args.Name = naming.BrokerExchangeName(b, true)
 	dlxExchange, err := r.rabbit.ReconcileExchange(ctx, args)
```
ikvmw, line 279: it's not clear that dlxExchange is not an Exchange anymore but a Result.

maybe other vars as well

ikvmw (changes requested):

pkg/reconciler/broker/broker.go
benmoss: It still does represent the exchange, in that the `Ready` and `Name` fields are about the exchange, it's just a different struct than the CRD type. I don't really like calling the variable `result`, since that's just generic. I'm not crazy about `exchangeResult` either, I don't see a problem with it being called `exchange` still.

deadtrickster:
LGTM

deadtrickster (approved):

ikvmw:
LGTM

ikvmw (approved):

PR #579, opened 2022-01-20
Fixing workflow errors

gabo1208:
- removed k8s v1.20 cause is not compatible with knative 1.1.0+
- added working directory to failure conformance workflows step


ikvmw:
From Slack:

Why adding working dir was needed?

Cause the gather failure step was calling make env outside the root repo dir, that was causing the "no target .env when calling make"

Issue #581, opened 2022-01-21
Define schema for rabbitmqsources.sources.knative.dev


ChunyiLyu:
**Problem**

No actual schema is defined for `rabbitmqsources.sources.knative.dev` at the moment. CRD has `x-kubernetes-preserve-unknown-fields` set to true. See: https://github.com/knative-sandbox/eventing-rabbitmq/blob/main/config/source/300-rabbitmqsource.yaml#L41

**Exit Criteria**

After installing `rabbitmqsources.sources.knative.dev` CRD, I can see information about its schema when I run `kubectl describe crd rabbitmqsources.sources.knative.dev`

**Time Estimate (optional):**

1 - 2?

**Additional context (optional)**
Add any other context about the feature request here.

ChunyiLyu:
As first step, I created PR: https://github.com/knative-sandbox/eventing-rabbitmq/pull/586 to 1) remove openapi-gen annotation 2) use the schema tool and generate a schema

That's not all. We have plenty of fields that uses format `_` rather than camel casing which is not to k8s convention (e.g. https://github.com/knative-sandbox/eventing-rabbitmq/blob/main/pkg/apis/sources/v1alpha1/rabbitmq_types.go#L74). Field names should be reconsidered as well (e.g. https://github.com/knative-sandbox/eventing-rabbitmq/blob/main/pkg/apis/sources/v1alpha1/rabbitmq_types.go#L99 why not name it autoDelete? Which is a known rabbitmq property). In addition to that, I think we should look into whether `rabbitmqsources` queueConfig and exchangeConfig could directly imbed queue spec and exchange spec from topology operator.

ChunyiLyu:
**Investigation on using controller-gen**

I tried using controller to generate rabbitmqsource crd. It does not work because of `spec.sink` which is type `*duckv1.Destination` and `status` which is `duckv1.SourceStatus`. Reason being both of them have `*apis.URL` as one of their fields. `apis.URL` is a knative type alias on `url.URL`. `url.URL` is a struct with no JSON tag specified, so controller-gen cannot successfully parse it.

Related issues reported from others in knative community:
- https://github.com/kubernetes-sigs/controller-tools/issues/625

Error from running controller gen:
```
controller-gen crd paths="./pkg/apis/sources/v1alpha1/" output:crd:artifacts:config=config/source
[...]

PR #586, opened 2022-01-25
Rabbitmqsources crd schema

ChunyiLyu:
# Changes

- use the schema tool to generate CRD (https://github.com/knative/hack/tree/main/schema); this tool has some issues. It cannot find field description successfully and pulls incorrect fields as `required`, some manual changes were required
- :bug: set `rabbitmqsources.spec.topic` to be optional (I saw from the new source example and topic could be set to empty string, this could be an optional field: https://github.com/knative-sandbox/eventing-rabbitmq/blob/main/samples/source/500-source.yaml#L8)
- :broom: remove openapi-gen annotation because it's not used

Related to #581

#581 requires more work, this PR shouldn't close the issue.  See [comment](https://github.com/knative-sandbox/eventing-rabbitmq/issues/581#issuecomment-1021189228) for suggested followups.

**Release Note**

Define schema for `rabbitmqsources`

**Docs**


xtreme-sameer-vohra:
@ChunyiLyu would you be able to update the state of the PR to deal with the conflicts ?

ChunyiLyu:
Closing this PR since we could use controller gen to generate rabbitmqsource after merging this: https://github.com/knative-sandbox/eventing-rabbitmq/pull/613

Issue #588, opened 2022-01-25
Fix Conformance testing


ikvmw:
**Describe the bug**
As of now I see that conformance test likely not pass on GH and locally inside kind

 **Expected behavior**
Robust conformance tests

 **To Reproduce**
`make test-conformance`

**See also**
https://github.com/knative-sandbox/eventing-rabbitmq/runs/4909441208?check_suite_focus=true#step:6:2798
https://github.com/knative-sandbox/eventing-rabbitmq/runs/4888976730?check_suite_focus=true#step:6:2869

ikvmw:
What we have found with @gsatchi 1) There is potentially a huge resource leak while running e2e and/or conformance tests via make locally 2) since https://github.com/knative-sandbox/eventing-rabbitmq/pull/567 master doesn't flake but continuously and reliably red. The PR was red itself. Basically our process allowed our team to merge red PRs. I assume that it happened because conformance tests are known to be flaky and well the usual outcome - ignoring. EDIT: link to our commits, note build status: https://github.com/knative-sandbox/eventing-rabbitmq/commits/main

xtreme-sameer-vohra:
Closing this as we're going to explore individual failures separately

PR #598, opened 2022-01-31
Fixing conformance and e2e workflows tests

gabo1208:
Improving the situación around #588

I approached it seems that each test environment was growing to big for it to be stable. So first I encounter some timeouts on the make install while running it locally that went away by bumping up the versions of some of the operatos.
Then manager to found that the controlplane conformance test was failing always locally to so I split it up.
Finally there was a flaky e2e test was setting up a failer service before the broker and was having issues when the failer stwrted sending messages and the broker was not up yep.

# Changes
- decreasing parallelism in conformance test
- increasing timeouts in sensible tests
- Fixing posible test conditions in e2e test using the fáciles service
- removing extra kind worker note cause it was causing failures in e2e tests (this is a theorem I haven't tested it with the currently setup)
- bumped up dependencias versiones in the makefile


Makefile
```diff
@@ -313,7 +313,7 @@ install-rabbitmq-cluster-operator: | $(KUBECONFIG) $(KUBECTL) ## Install RabbitM
 # https://github.com/jetstack/cert-manager/releases
 # ⚠️  You may want to keep this in sync with RABBITMQ_TOPOLOGY_OPERATOR_VERSION
 # In other words, don't upgrade cert-manager to a version that was released AFTER RABBITMQ_TOPOLOGY_OPERATOR_VERSION
```
gab-satchi, line 315: The RABBITMQ_TOPOLOGY_OPERATOR_VERSION is from Oct 2021. Should we also update that to the latest? (1.3.0) Just thinking about the warning above with keeping the versions in-sync

gab-satchi:
hey @gabo1208, seeing some `Message:0/2 nodes are available: 1 Insufficient cpu, 1 node(s) had taint {node-role.kubernetes.io/master: }, that the pod didn't tolerate.` in the conformance test failures

test/conformance/conformance_test.go
```diff
@@ -52,7 +52,46 @@ func TestBrokerConformance(t *testing.T) {
 	env.Prerequisite(ctx, t, secret.Created("rabbitbroker-secret", "rabbitbroker", ctx, b.WithEnvConfig()...))
 	// Install and wait for a Ready Broker.
 	env.Prerequisite(ctx, t, broker.GoesReady("default", b.WithEnvConfig()...))
+	brokerFeatureSet := broker.ControlPlaneConformance("default")
+	// To distribute the test total load, 8 gave me the best results locally
+	brokerFeatureSet.Features = brokerFeatureSet.Features[:8]
+	env.TestSet(ctx, t, brokerFeatureSet)
+}
+
+func TestBrokerControlPlaneConformance2(t *testing.T) {
```
gab-satchi, line 61: I understand setting `-parallel=8` to lighten the load but I'm confused about splitting up the Conformance tests by feature here. I don't know much abt the reconciler-test framework that sets up the features, but I naively thought setting the parallel flag would be enough?

gabo1208: I tested locally and without the split up the errors appeared

gabo1208: What I think can be happening, is that the env of the feature set is not been cleaned up  when its in the same test suite. But splitting test suites the feature set env get cleaned up more times

gab-satchi: Yeah it looks like it would recreate the env if split into different tests. From what I can tell, the env is pretty much a RabbitMQ cluster and a default broker? I don't think the default broker gets used, it feels more like a way to get a ready signal before kicking off the tests. The tests seem to create brokers for themselves

gabo1208: Yep, that's what the basic tests create, buuut. In the first test the broker is just been tested for the ready condition, in the other tests sometimes triggers are being created for the default broker and sometimes for a new broker. Is pretty weird. To give more context, there are a total of 18 features that are being tested

gab-satchi: Ah yes I see it, thanks. Some seem to use the randomized name and then some use the broker given to them.

gabo1208: It's getting better at least xD just one Red workflow!

PR #602, opened 2022-02-03
Adds policyInformer to Trigger and Broker controllers

gab-satchi:
Observed a state where a Trigger can be in an unready state even though the Queue and Policy are both in Ready. 

# Changes
The trigger controller will now watch for events in policies and enqueue the trigger if it's the owner of the policy.
The broker controller will also watch for policies owned by Brokers

/kind bug

Addresses some of the flakes observed in #588

### DIFF
--- a/test/e2e/kind.yaml
+++ b/test/e2e/kind.yaml
@@ -19,5 +19,3 @@ nodes:
   image: kindest/node:${KIND_K8S_VERSION}@${KIND_K8S_DIGEST}
 - role: worker
   image: kindest/node:${KIND_K8S_VERSION}@${KIND_K8S_DIGEST}
-- role: worker
-  image: kindest/node:${KIND_K8S_VERSION}@${KIND_K8S_DIGEST}
--- a/test/e2e/kind.yaml
+++ b/test/e2e/kind.yaml
@@ -19,3 +19,5 @@ nodes:
   image: kindest/node:${KIND_K8S_VERSION}@${KIND_K8S_DIGEST}
 - role: worker
   image: kindest/node:${KIND_K8S_VERSION}@${KIND_K8S_DIGEST}
+- role: worker
+  image: kindest/node:${KIND_K8S_VERSION}@${KIND_K8S_DIGEST}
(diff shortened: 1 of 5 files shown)


test/e2e/kind.yaml
```diff
@@ -19,3 +19,5 @@ nodes:
   image: kindest/node:${KIND_K8S_VERSION}@${KIND_K8S_DIGEST}
 - role: worker
   image: kindest/node:${KIND_K8S_VERSION}@${KIND_K8S_DIGEST}
+- role: worker
+  image: kindest/node:${KIND_K8S_VERSION}@${KIND_K8S_DIGEST}
```
gab-satchi, line 23: Temporary till we get a signal with these changes.

gab-satchi:
This change seems to consistently get passed the `Setup` phase but there is now a flake that could happen in the `Requirement` phase. Opened #603 to address it

ikvmw (changes requested):
I and @gabo1208 looked it at and we have a question. Informers set up handlers to enqueue events, but where is the actual processing is done? Like we see only producing part, where is Policy changes processed.

gab-satchi:
> I and @gabo1208 looked it at and we have a question. Informers set up handlers to enqueue events, but where is the actual processing is done? Like we see only producing part, where is Policy changes processed.

This repo doesn't reconcile the Policy, that's done by the topology-operator. We do however rely on the Policy's status in order to set the Trigger's status. The handler will trigger a reconcile on the Trigger/Broker when a policy they own is updated. 

Any Policy changes will get reflected onto the Trigger/Broker through this [codeblock](https://github.com/knative-sandbox/eventing-rabbitmq/blob/00e9d0ca8889ca635fdd95899ad8f753481db2b0/pkg/rabbit/service.go#L94-L117)

ikvmw (approved):

ikvmw:
Per zoom: Informer's handler enqueues event that will be later picked up by controller loop: https://github.com/knative-sandbox/eventing-rabbitmq/blob/56717fb27f1980c4306b9988bc570cd1f919e3a8/vendor/knative.dev/pkg/controller/controller.go#L493.

Which shows us that knative uses its own controller framework and not the one from Operator SDK - https://github.com/kubernetes-sigs/controller-runtime

Issue #550, opened 2021-12-13
Figure out why CodeCov fails on trivial changes


benmoss:
Why is the CodeCov action failing here? https://github.com/knative-sandbox/eventing-rabbitmq/pull/530

PR #605, opened 2022-02-08
Switch to using DeepEqual so changes don't get missed

gab-satchi:
# Changes
When reconciling the dispatcher and DLX dispatcher deployments for a Trigger, use DeepEqual. This is to fix the following edge case:
- Create a Trigger with the `rabbitmq.eventing.knative.dev/prefetchCount: "10"` annotation. 
- Wait for the deployments to get created with PREFETCH_COUNT env variable set to 10
- Edit the Trigger and delete the annotation

Expected: 
- Deployment is updated and a new Pod is created with the default PREFETCH_COUNT of 1

Actual:
- Deployment (and Pod) aren't getting updated.

This happens because we've been using `reflect.DeepDerivative` which will ignore any unset fields in the first argument. Switching to `reflect.DeepEqual` fixes the issue.

- 🐛  Fix edge case with Trigger deployment reconciliation

/kind bug


gabo1208:
A test case for this would be nice, and let's start using hold to wait for at least 2 approvals in the PR's 🙂  cc @ikvmw @ChunyiLyu @xtreme-sameer-vohra @embano1

gab-satchi:
let's continue to hold for a test case. will send an update

ChunyiLyu (approved):
LGTM

xtreme-sameer-vohra:
Is this related https://github.com/knative-sandbox/eventing-rabbitmq/issues/550

gabo1208 (approved):

PR #608, opened 2022-02-16
[Docs] Adds trigger performance examples

### LINKED ISSUE #424: Document a Trigger's Behavior for FIFO, High Throughput modes
Ensure the configuration and expected behavior is documented for end users, specifically
- message channel is rate-limited by the prefetch count, so the number of goroutines will equal the prefetch
- setting prefetch to 1 with no autoscaling would be equivalent of FIFO

gab-satchi:
# Changes
Starts a `trigger-customization` section and adds prefetch count examples

/kind documentation

Fixes #424

### DIFF
--- a/samples/trigger-customizations/500-sink.yaml
+++ b/samples/trigger-customizations/500-sink.yaml
@@ -1,0 +1,25 @@
+apiVersion: serving.knative.dev/v1
+kind: Service
+metadata:
+  name: fifo-event-display
+  namespace: trigger-demo
+spec:
+  template:
+    spec:
+      containers:
+      - image: ko://knative.dev/eventing-rabbitmq/samples/trigger-customizations/event-display
+        args:
+          - --delay=20
+---
+apiVersion: serving.knative.dev/v1
+kind: Service
+metadata:
+  name: high-throughput-event-display
+  namespace: trigger-demo
+spec:
+  template:
+    spec:
+      containers:
+        - image: ko://knative.dev/eventing-rabbitmq/samples/trigger-customizations/event-display
+          args:
+            - --delay=20
--- a/samples/trigger-customizations/500-sink.yaml
+++ b/samples/trigger-customizations/500-sink.yaml
@@ -5,6 +5,9 @@ metadata:
   namespace: trigger-demo
 spec:
   template:
+    metadata:
+      annotations:
+        autoscaling.knative.dev/min-scale: "1"
     spec:
       containers:
       - image: ko://knative.dev/eventing-rabbitmq/samples/trigger-customizations/event-display
@@ -18,6 +21,9 @@ metadata:
   namespace: trigger-demo
 spec:
   template:
+    metadata:
+      annotations:
+        autoscaling.knative.dev/min-scale: "1"
     spec:
       containers:
         - image: ko://knative.dev/eventing-rabbitmq/samples/trigger-customizations/event-display
--- a/samples/trigger-customizations/600-event-sender.yaml
+++ b/samples/trigger-customizations/600-event-sender.yaml
@@ -1,0 +1,14 @@
+apiVersion: v1
+kind: Pod
+metadata:
+  name: event-sender
+  namespace: trigger-demo
+spec:
+  restartPolicy: Never
+  containers:
+    - name: event-sender
+      image: ko://knative.dev/eventing-rabbitmq/samples/trigger-customizations/event-sender
+      args:
+        - --sink=http://default-broker-ingress.trigger-demo.svc.cluster.local
+        - --event={"specversion":"1.0","type":"SenderEvent","source":"Sender","datacontenttype":"application/json","data":{}}
+        - --num-messages=10
--- a/samples/trigger-customizations/README.md
+++ b/samples/trigger-customizations/README.md
@@ -7,19 +7,6 @@
 - Set `KO_DOCKER_REPO` to an accessible container registry
 
 ## Overview
-
-This doc and demos highlight the various customizations that can be done on a Trigger along with their usecases and implications.
-
-### Trigger Pre-Fetch Count
-Trigger has a configurable annotation `rabbitmq.eventing.knative.dev/prefetchCount`. The following are effects of setting this parameter to `n`:
-
-- Prefetch count is set on the RabbitMQ channel and queue created for this trigger. The channel will receive a maximum of `n` number of messages at once.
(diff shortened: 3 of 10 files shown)


gabo1208:
There are some trailing white spaces alert from reviewdog that need to be fixed, apart from that, everything is looking fine

samples/trigger-customizations/600-event-sender.yaml
```diff
@@ -0,0 +1,14 @@
+apiVersion: v1
+kind: Pod
+metadata:
+  name: event-sender
+  namespace: trigger-demo
+spec:
+  restartPolicy: Never
+  containers:
+    - name: event-sender
+      image: ko://knative.dev/eventing-rabbitmq/samples/trigger-customizations/event-sender
```
gabo1208, line 10: Same comment here with the [ping source](https://github.com/knative/eventing/blob/main/config/core/resources/pingsource.yaml) or the [Container Source](https://knative.dev/docs/eventing/custom-event-source/containersource/)

samples/trigger-customizations/500-sink.yaml
```diff
@@ -0,0 +1,25 @@
+apiVersion: serving.knative.dev/v1
+kind: Service
+metadata:
+  name: fifo-event-display
+  namespace: trigger-demo
+spec:
+  template:
+    spec:
+      containers:
+      - image: ko://knative.dev/eventing-rabbitmq/samples/trigger-customizations/event-display
```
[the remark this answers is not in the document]
gab-satchi: The main thing missing from the other display type receivers I came across was the delay. I wanted to mimic a slow receiver by having it wait for x seconds before responding to the event

samples/trigger-customizations/600-event-sender.yaml
gab-satchi: For the sender, I looked at the test-images provided by eventing-core but images like the heartbeat required a period > 0. Ping source relies on a cron expression as well. So they send a message per x seconds. I needed something that would spam x messages as quickly as possible.

gabo1208: With the container source you can set ms too (just In case) but I'm approving this :)

gab-satchi: If we can avoid maintaining another test_image I'm all for it. I saw the container source using the heartbeats test_image as well n was deterred by [this](https://github.com/knative/eventing/blob/main/test/test_images/heartbeats/main.go#L107-L112). Maybe I can modify that to take a string for duration instead like `5s` or `5n` which will help us reuse it

gab-satchi: Was in the middle of making the switch then I realized with heartbeats, you can't limit the number of messages. Ideally I want something like 10 messages or so. With creating heartbeat container and setting a low period like 1ms, it becomes hard to observe events at the display with all of the messages coming in very quickly without and end.

samples/trigger-customizations/500-sink.yaml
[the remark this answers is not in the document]
xtreme-sameer-vohra: Could it be done with the vanilla event-display with setting https://knative.dev/docs/serving/autoscaling/concurrency/#hard-limit to 1 and https://knative.dev/docs/serving/autoscaling/scale-bounds/#lower-bound to > 1 for the high throughput case ?

samples/trigger-customizations/600-event-sender.yaml
gabo1208: In this case, I'd rather to explain something like logs -f to the user than to have another image to maintain. But I'll leave it up to you to decide. Either way, I'm approving :)

samples/trigger-customizations/README.md
```diff
@@ -0,0 +1,265 @@
...
+NOTE: ko is used here to create the custom images. Ensure `KO_DOCKER_REPO` is set to an accessible repository for the images.
+```sh
+ko apply -f samples/trigger-customizations/500-sink.yaml
+```
+or
+```sh
+ko apply -f - << EOF
+apiVersion: serving.knative.dev/v1
+kind: Service
+metadata:
```
xtreme-sameer-vohra, line 149: Recommend setting minScale to 1 - https://knative.dev/docs/serving/autoscaling/scale-bounds/#lower-bound just to prevent the pod from disappearing during the demo

samples/trigger-customizations/README.md
```diff
@@ -0,0 +1,265 @@
...
+  template:
+    spec:
+      containers:
+      - image: ko://knative.dev/eventing-rabbitmq/samples/trigger-customizations/event-display
+        args:
+          - --delay=20
+---
+apiVersion: serving.knative.dev/v1
+kind: Service
+metadata:
```
xtreme-sameer-vohra, line 162: Same as above

xtreme-sameer-vohra (commented):
Looks good. Added some comments.

samples/trigger-customizations/README.md
```diff
@@ -0,0 +1,265 @@
...
+
+### Trigger Pre-Fetch Count
+Trigger has a configurable annotation `rabbitmq.eventing.knative.dev/prefetchCount`. The following are effects of setting this parameter to `n`:
+
+- Prefetch count is set on the RabbitMQ channel and queue created for this trigger. The channel will receive a maximum of `n` number of messages at once.
+- The trigger will create `n` workers to consume messages off the queue and dispatch to the sink.
+
+If this value is unset, it will default to `1`. This means the trigger will only handle one event at a time. This will preserve the order of the messages in a queue but
+will make the trigger a bottleneck. A slow processing sink will result in low overall throughput. Setting a value higher than 1 will result in `n` events being handled at
+a time by the trigger but ordering won't be guaranteed as events are sent to the sink.
```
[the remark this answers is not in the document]
gab-satchi: Did you mean moving the full README under there as a separate doc or moving this block over to broker/README?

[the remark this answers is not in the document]
xtreme-sameer-vohra: The latter. Mostly to keep the documentation in one place and the samples could refer to the documentation 🤔

xtreme-sameer-vohra:
/lgtm
The only outstanding minor change can be done in a separate PR as well.

samples/trigger-customizations/README.md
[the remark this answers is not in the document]
gab-satchi: updated. thanks

Issue #504, opened 2021-11-10
Source should use some kind of backoff


benmoss:
I have the source installed and it's failing to deliver messages to its subscriber and it's currently warming up my apartment with its retry rate 😄
![Screen Shot 2021-11-10 at 4 43 47 PM](https://user-images.githubusercontent.com/239754/141198473-03f62b28-d1c0-4ea5-a393-46c36daa153e.png)
 

# Open Question
- Is this change going to be additive or breaking the API ? If its the former and doesn't break the API, lets tackle this post MVP.

gabo1208:
Closed by #614

PR #614, opened 2022-02-18
Source backoff strategy on failed events

### LINKED ISSUE #537: There is no failover or protection when events fail to be sent from the RabbitMQ Source
**Describe the bug**
When there is an error sending the events from the Source to a Sink they just keeps getting reqeued and RabbitMQ goes crazy:

screenshot_2021-12-02_at_15 58 22

When the peak in message rate in the graph starts to go down is when we configured the Knative DNS in the Cluster.

 **Expected behavior**
When message are not been sent, that the Source stops and wait for the Sink to be available. And maybe that the source (if the sink is a Ref) checks that a resolvable Ref.

 **To Reproduce**
- Spun up a Cluster environment with knative, but do not configure the DNS
- Create a RabbitMQ cluster
- Create a Source with a sink (uri or ref)
- Create the sink (optional)
- Get to the RabbitMQ Management UI and see how the image above gets replicated

 **Knative release version**
1.0.0

gabo1208:
# Changes

- :gift: Now using CE http retries when message fails
- :gift: There is a backoff strategy when retrying
- 🧹 Not reqeueing when prefetchCount is 1 (FIFO)

/kind performance

Fixes #537 #504

**Release Note**

```release-note
Now the RabbitMQ Source, when failing to send an event, does not get stuck in a cycle requeueing messages as they come. It follows a retries based backoff strategy (and does not reqeue them if the prefetchCount is 1)
```


pkg/apis/sources/v1alpha1/rabbitmq_types.go
```diff
@@ -124,6 +125,22 @@ type RabbitmqSourceSpec struct {
 	// RabbitMQ cluster
 	// +optional
 	Predeclared bool `json:"predeclared,omitempty"`
+	// Retry is the minimum number of retries the sender should attempt when
+	// sending an event before moving it to the dead letter sink.
+	// +optional
+	Retry *int32 `json:"retry,omitempty"`
```
xtreme-sameer-vohra, line 131: Would this and `BackoffPolicy`,  `BackoffDelay` be available for users to configure or they'd have fixed values in code but wouldn't be configurable by users ?

gabo1208: After what we talked about today, I think this makes sense (they are configurable by users but have a default). Also the Eventing Source Spec "explicitly permit" additional data in the Source's Spec and Status (https://github.com/knative/eventing/blob/main/docs/spec/sources.md#source-custom-objects)

source/source-example.yaml
```diff
@@ -6,6 +6,9 @@ metadata:
 spec:
   brokers: RABBITMQ_BROKERS
   topic: RABBITMQ_TOPIC
+  retry: 5
+  backoffPolicy: "linear"
+  backoffDelay: "PT30S"
```
gab-satchi, line 11: Does the period format "PT30S" work with time.Duration?

gab-satchi (commented):
mostly looks good to me. just a few nits.

source/source-example.yaml
gabo1208: You are right let me change it

gab-satchi (commented):
/lgtm

gab-satchi:
@gabo1208 maybe a squash to clean up the commits a bit :)

PR #628, opened 2022-03-08
Patch to get conformance tests running for the kn release v1.3

gabo1208:
# Changes
- removed resource limitation from the rabbitmq cluster yaml def
- added 2 more worker nodes to the kind cluster
- reorganized conformance tests
- increased conformance tests timeout


test/e2e/kind.yaml
```diff
@@ -17,6 +17,10 @@ kubeadmConfigPatches:
 nodes:
 - role: control-plane
   image: kindest/node:${KIND_K8S_VERSION}@${KIND_K8S_DIGEST}
+- role: control-plane
```
gab-satchi, line 20: did you mean for this to be a worker as well? Can't have 2 control planes. I also don't think scaling control planes would add much to our testing.

gabo1208: Yeah I will correct this in my next commit 👍🏼

PR #660, opened 2022-03-21
Increase memory request/limit to 1Gi

gab-satchi:
- 🐛  increase memory request to 1Gi for RabbitMQ clusters created during conformance testing

/kind cleanup

The tests occasionally run into an issue where during startup, the RabbitMQ server is using too much memory (~160Mi) and that triggers the memory alarm threshold. When that alarm is triggered, producers are blocked. This causes a failure in the test where the message is stuck at the ingress pod and doesn't proceed any further.


ChunyiLyu (approved):
LGMT

PR #662, opened 2022-03-24
Create messaging topology objects for rabbitmqsource

### LINKED ISSUE #6: Source's RMQ resources aren't cleaned up on delete
**Describe the bug**
As far as I can tell, in the current implementation, even if a Source resource is deleted, RMQ resources will remain (which means the Queue continues to receive messages via the Binding).

 **Expected behavior**
RMQ resources are cleaned up on deleting a Source
- binding (always)
- queue (if [spec.predeclared](https://github.com/knative-sandbox/eventing-rabbitmq/tree/main/source#creating-and-managing-sources) is set to false)

 **Open Questions** 
- Should the exchange be deleted as well ?

 **To Reproduce**

 **Knative release version**

--------------------
Original Issue body
--------------------
> Don't you also need to have FinalizeKind for any cleanup? Just jotting this down so I don't forget :)
> 
> _Originally posted by @vaikas in https://github.com/knative-sandbox/eventing-rabbitmq/pull/3#discussion_r427597886
> 
> well, there will also typically be resources created on the RabbitMQ cluster as well (Exchange and Queue, unless they were pre-existing) and the Binding from the Exchange to the Queue. Those need to be cleaned up on > deletion as well, and the Binding might need to be replaced if there is an update to source properties that affect it
> 
> _Originally posted by @markfisher in https://github.com/knative-sandbox/eventing-rabbitmq/pull/3#discussion_r427621633
> 
> Yeah, that's why I tagged it first above since I was expecting there to be some sort of cleanup necessary, but I must have missed (as I said, just the first pass) on what other cleanup needs to happen. Maybe you could > add a comment in the PR at the location where the creation of resources happens that need to get cleaned up, so it won't get lost?
> 
> _Originally posted by @vaikas in https://github.com/knative-sandbox/eventing-rabbitmq/pull/3#discussion_r428778848
> 
> Perhaps the lifecycle of these resources (Exchange, Queue and the Bindings below) should be managed by the reconciler? That would allow for updates/deletes to occur there as well. This code could then be changed to use > "passive" versions of the declare operations.
> 
> As far as I can tell, in the current implementation, even if a Source resource is deleted, these resources will remain (which means the Queue continues to receive messages via the Binding).
> 
> _Originally posted by @markfisher in https://github.com/knative-sandbox/eventing-rabbitmq/pull/3#discussion_r428795351

### LINKED ISSUE #528: Deleting a RabbitMQ Source should delete the queue
**Describe the bug**
When a `RabbitmqSource` is deleted, the queue that it creates is left behind.
This was observed part of https://github.com/knative-sandbox/eventing-rabbitmq/pull/522#issuecomment-981820162

 **Expected behavior**
When the `RabbitmqSource` is deleted, the queue should be deleted too, otherwise it will continue receiving messages:
image

 **To Reproduce**
Deploy the sample added in https://github.com/knative-sandbox/eventing-rabbitmq/pull/522. Delete the `RabbitmqSource` with the following config and observe the queue remaining behind.

```yaml
---
kind: RabbitmqSource
metadata:
  name: rabbitmq-source
  namespace: concurrent-dispatch
spec:
  brokers: "rabbitmq:5672/"
  topic: ""
  user:
    secretKeyRef:
      name: rabbitmq-default-user
      key: username
  password:
    secretKeyRef:
      name: rabbitmq-default-user
      key: password
  channel_config:
    prefetch_count: 10
    global_qos: false
  exchange_config:
    name: "eventing-rabbitmq-source"
    type: "fanout"
    durable: false
    auto_deleted: false
    internal: false
    nowait: false
  queue_config:
    name: "eventing-rabbitmq-source"
    routing_key: ""
    durable: false
    # TODO: should the queue be deleted when the source gets deleted?
    delete_when_unused: true
    exclusive: false
    nowait: false
  sink:
    ref:
      apiVersion: serving.knative.dev/v1
      kind: Service
      name: rabbitmq-source-sink
      namespace: concurrent-dispatch
```

 **Knative release version**
Tested with v1.0.0, eventing-rabbitmq dev version (#522) all running on GKE v1.21.5.

 **Additional context**
@gabo1208 shares the context.

It was observed that using the `Exclusive` flag results in the correct behaviour, so my suspicion is that this is a RabbitMQ issue.

ChunyiLyu:
# Changes

- :gift: rabbitmqsource controller used to create queues, exchanges and
bindings via connecting to rabbit server directly. This PR switch to use messaging topology operator to create rabbit resources if they are not predeclared.
- :bug: owner reference is set on created queues, exchanges, and bindings
which means they will be deleted when rabbitmqsource is deleted

/kind api-change
/kind bug

Fixes #528 
Fixes #6


gabo1208:
Everything ok so far, maybe it should update the samples? Or you prefer to do that in a separate PR?

ChunyiLyu:
@gabo1208 I'm still working on solving the e2e and conformance failures. Will let you know when the PR is ready for a final review. Good catch on the examples! I will update them as part of this PR.

PR #670, opened 2022-03-31
Rename rmqsource exchangeConfig TypeOf to Type

ChunyiLyu:
# Changes

- :broom: This only renames the the struct field name, json key is 'type' already

/kind cleanup


gab-satchi (approved):
/lgtm

PR #673, opened 2022-03-31
Move NewPolicy() NewExchange() NewBinding() NewQueue() to pkg/rabbit

ChunyiLyu:
# Changes

- :broom: backfill test for CreateSourceRabbitName()
- 🧹 Move NewPolicy() NewExchange() NewBinding() NewQueue() to `pkg/rabbit`. These function were only used in `/pkg/rabbit` and it's rather messy for them to live in separate places in`pkg/reconciler/trigger` and `pkg/reconciler/broker`

/kind cleanup

### DIFF
--- a/pkg/rabbit/binding.go
+++ b/pkg/rabbit/binding.go
@@ -14,7 +14,7 @@ See the License for the specific language governing permissions and
 limitations under the License.
 */
 
-package resources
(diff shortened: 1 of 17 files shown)


gab-satchi (approved):
/lgtm

PR #679, opened 2022-04-01
Name queue&exchange name required for rabbitmqsource

ChunyiLyu:
- :gift: - queue name is always required; exchange name is required when spec.predeclared is false

/kind enhancement


gab-satchi (approved):
Thanks for the changes!

/lgtm

PR #680, opened 2022-04-01
Add connectionSecret to rabbitmqsource examples

ChunyiLyu:
# Changes

- 📜  Add connectionSecret to rabbitmqsource examples

/kind documentation

### DIFF
--- a/samples/source/500-source.yaml
+++ b/samples/source/500-source.yaml
@@ -14,6 +14,8 @@ spec:
     secretKeyRef:
       name: rabbitmq-default-user
       key: password
+  connectionSecret:
+    name: rabbitmq-default-user # used when exchange and queue are not predeclared
   channelConfig:
     # Value must be between 1 and 1000
     # A value of 1 RabbitMQ Source behaves as a FIFO queue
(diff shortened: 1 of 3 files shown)


samples/source/500-source.yaml
```diff
@@ -14,6 +14,8 @@ spec:
     secretKeyRef:
       name: rabbitmq-default-user
       key: password
+  connectionSecret:
```
gab-satchi, line 17: Should [this](https://github.com/knative-sandbox/eventing-rabbitmq/tree/main/samples/source#create-the-rabbitmq-source) also be updated to reflect the connectionSecret?

gab-satchi (commented):
wondering if connectionSecret should be referenced elsewhere in the docs

samples/source/500-source.yaml
ChunyiLyu: You are totally right. I should update the step to add key `uri` to the connection secret as well.

gab-satchi (approved):
/lgtm

gabo1208 (approved):

PR #687, opened 2022-04-04
Adds timeout to dispatcher.

### LINKED ISSUE #671: Trigger's sink takes longer than X to respond (are there any timeouts on the dispatcher acting as a client? Lets ensure there is a defined timeout rather than no timeout - Reference)
The current behaviour is using the ceClient to send a POST HTTP request, and using net/http Client which doesn't define any timeout to start with.

Some reference on how this can be set up: https://blog.cloudflare.com/the-complete-guide-to-golang-net-http-timeouts/ 
Old references: https://medium.com/@nate510/don-t-use-go-s-default-http-client-4804cb19f779 

In my opinion, we should have a configuration to define this timeout on the client inside the dispatcher.

gab-satchi:
- hardcoded timeout of 30s
- modified tests

# Changes
- 🎁  Adds timeout for dispatcher communication with subscriber

NOTE: this change adds about 30s to our unit tests as the timeout is currently not configurable. It didn't make sense to make it configurable just for the tests. 
If we choose to make it configurable to the user, we could use the same mechanism to start the dispatcher with lower timeouts for testing.

/kind enhancement

Fixes #671 

```release-note
- dispatcher will timeout after 30s when sending to the subscriber
```


samples/trigger-customizations/event-display/main.go
```diff
@@ -42,6 +42,7 @@ func display(event cloudevents.Event) {
 }
 
 func main() {
+	flag.Parse()
```
gab-satchi, line 45: this change is unrelated to this PR. Just something I saw was missing while testing out changes.

salaboy:
@gab-satchi we discussed this feature in yesterday's Eventing WG, and because the Kafka folks already have this implemented, they are thinking to promote the feature out of experimental and even pushing some changes to the spec (I haven't reviewed that yet). I believe this is a great way to start engaging with upstream, so we can follow the change all the way up to the spec.

gab-satchi:
I think it will be great to add this to the upstream spec. It may make sense to have this default timeout in place as a stopgap to prevent connections being held forever at the dispatcher. 

Once the eventing core work is done, we can modify this to check there for a timeout and fallback to the default.

lionelvillard:
There is a PR to add this to the spec: https://github.com/knative/specs/pull/76.

gabo1208:
Is this ready to review cc @gab-satchi ?

gab-satchi:
yes it's ready @gabo1208. the 30s timeout will be stopgap while we work on making it configurable in the spec

gab-satchi:
/hold 

will update to use the experimental feature in deliverySpec

gabo1208 (approved):

PR #700, opened 2022-04-08
Cleanup code

ChunyiLyu:
# Changes

- :broom: extracted a common function to set rabbitmq objects labels and some small code clean up

/kind cleanup


pkg/rabbit/label.go
```diff
@@ -0,0 +1,48 @@
+/*
+Copyright 2020 The Knative Authors
```
gab-satchi, line 2: nit: should be 2022

pkg/rabbit/label.go
```diff
@@ -0,0 +1,48 @@
...
+
+const (
+	TriggerLabelKey = "eventing.knative.dev/trigger"
+	SourceLabelKey  = "eventing.knative.dev/SourceName"
+)
+
+// Labels generates the labels for a RabbitMQ resource
+// Used by exchanges, queues, and bindings created by broker, trigger, and source controllers
+func Labels(b *eventingv1.Broker, t *eventingv1.Trigger, s *v1alpha1.RabbitmqSource) map[string]string {
+	if t != nil {
```
gab-satchi, line 33: Currently there's an implicit expectation that if trigger isn't nil then broker shouldn't be nil either. Maybe swap this to a: `t != nil && b!= nil` to avoid a nil exception at L35

gab-satchi (commented):
just some minor things

pkg/rabbit/label.go
ChunyiLyu: Corrected :)

pkg/rabbit/label_test.go
```diff
@@ -0,0 +1,87 @@
+/*
+Copyright 2020 The Knative Authors
```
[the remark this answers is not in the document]
ChunyiLyu: Corrected :)

pkg/rabbit/label.go
ChunyiLyu: Good catched! Fixed now

gab-satchi (approved):
/lgtm

PR #701, opened 2022-04-08
Set default for rmqsource in CRD instead of webhook

ChunyiLyu:
# Changes

- Set default for rmqsource in CRD instead of webhook. It's cleaner and more k8s idiomatic this way.

/kind enhancement


gab-satchi (approved):
/lgtm

Issue #646, opened 2022-03-15
Create one policy for triggers without dead letter configurations 


ChunyiLyu:
**Current behavior**

When users create a trigger, if the trigger has a dead letter configured, we create a policy to configure the trigger queue with the provided dead letter exchange; when the trigger does not have a dead letter configured, we create a policy to configure the trigger queue with the broker dead letter exchange, so either way we are creating one policy per trigger.

Rabbit policies match queue names based on the provided regex. It's particularly useful when configuring queues in a group. However, the current broker trigger implementation creates one policy per queue, which is slightly of an anti pattern.

**Suggestions**

1. When rabbitmq broker is created, create a policy that uses a regex pattern for queue name matching with the broker dead letter exchange
2. If a trigger does not have a dead letter configured, we create the trigger with a queue name that matches the broker dead letter exchange policy regex
3. If a trigger has a dead letter configured, we can create a dedicated policy (updatable) or configure via queue argument (not updatable) to configure the trigger queue with the dedicated dead letter exchange

**[Persona:](https://github.com/knative/eventing/blob/main/docs/personas.md)**

**Exit Criteria**

**Time Estimate (optional):**

n/a

**Additional context (optional)**

ChunyiLyu:
@gab-satchi and I was discussing how the broker trigger creates policies with regards to the [dead letter e2e test flake](https://github.com/knative-sandbox/eventing-rabbitmq/issues/637).

Thoughts about this? @gabo1208 @ikvmw

ChunyiLyu:
@gabo1208 For triggers with dead letter queue defined, it will depends on whether we allow dead letter queue to be updated once the trigger is created. Queue arguments cannot be updated after the queue is created. If we want to allow dead letter queue to be updated, we will need to configure these trigger queues via policies (same as what we are currently doing). If  update is not something we'd like to support, we don't need to create a policy, we can use queue arguments to configure the trigger queue.

PR #714, opened 2022-04-13
Create one broker deadletter exchange policy

ChunyiLyu:
# Changes

<!-- 
Describe your changes here- ideally you can get that description straight from
your descriptive commit message(s)! 

- :gift: when broker has deadletter configured and trigger does not we create one broker deadletter policy, and create trigger queue with a specific prefix to use the broker dlx. This will reduce the number of policies created
- broker deadletter exchange policy is created with a priority lower than the trigger dlx policy, so trigger dlx could be later updated successfully
- contains some refactoring of the broker controller

/kind enhancement

closes: #646


pkg/rabbit/service.go
```diff
@@ -143,6 +143,32 @@ func (r *Rabbit) ReconcileBinding(ctx context.Context, args *BindingArgs) (Resul
 	}, nil
 }
 
+func (r *Rabbit) ReconcileBrokerDLXPolicy(ctx context.Context, args *QueueArgs) (Result, error) {
+	logging.FromContext(ctx).Infow("Reconciling broker dead letter policy", zap.String("name", args.Name))
```
gab-satchi, line 147: `ReconcileQueue` also reconciles a policy if the DLXName is defined. Does it make sense to make this a generic `ReconcilePolicy` and have the trigger controller use that as well?

gab-satchi: yep sure thing

gabo1208:
Why the hold?

pkg/rabbit/queue.go
```diff
@@ -79,10 +79,31 @@ func NewPolicy(args *QueueArgs) *rabbitv1beta1.Policy {
...
+		ObjectMeta: metav1.ObjectMeta{
+			Name:            args.Name,
+			Namespace:       args.Namespace,
+			OwnerReferences: []metav1.OwnerReference{args.Owner},
+			Labels:          args.Labels,
+		},
+		Spec: v1beta1.PolicySpec{
+			Name:                     args.Name,
+			Priority:                 0, // lower priority then policies created for trigger queues to allow overwrite
+			Pattern:                  fmt.Sprintf("^%s.", regexp.QuoteMeta("t.q.")),
```
gab-satchi, line 103: This pattern doesn't distinguish based on Broker name. If we have 2 brokers with delivery spec defined, it looks like they'll create two policies with identical patterns.

gab-satchi (commented):
I think we're going to have to introduce broker name as a prefix to trigger queue names in order to only target triggers that belong to a broker.

pkg/rabbit/queue.go
gabo1208: Oh ok ok, and avoid edge cases where in broker name is a subset of other

gab-satchi: Think the problem would exist even if one name wasn't the subset of the other. The pattern used for the policy is a static `^t.q.` which all trigger queues will fall under

gabo1208: Yeah I ment with the broker name prefix fix

pkg/reconciler/broker/broker_test.go
```diff
@@ -668,6 +738,7 @@ func TestReconcile(t *testing.T) {
 				createReadyExchange(true),
 				createReadyQueue(true),
 				createReadyBinding(true),
+				createReadyPolicy(),
```
gab-satchi, line 741: same thing here.

gab-satchi (commented):
looks good. just had one question around the tests

pkg/reconciler/broker/broker_test.go
```diff
@@ -630,6 +699,7 @@ func TestReconcile(t *testing.T) {
...
```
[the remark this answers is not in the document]
ChunyiLyu: Good point! Removed.

pkg/reconciler/broker/broker_test.go
ChunyiLyu: Good point! Removed.

ChunyiLyu:
@gab-satchi @gabo1208 Are we ok to merge this?

gab-satchi (commented):
/lgtm

gab-satchi:
/hold cancel

thanks for the changes Chunyi

PR #723, opened 2022-04-19
Broker and Trigger use delivery spec for dead letter messaging

gab-satchi:
- Plumbs deliverySpec.BackoffDelay through to the dispatcher

Note: The dispatcher creation for broker and trigger is very similar. Can be consolidated in a follow-up PR 

/kind bug

**Release Note**

```release-note
- backoffDelay is used when defined in Broker or Trigger delivery spec
- dead letter messaging uses the same delivery spec
```

### DIFF
--- a/pkg/reconciler/broker/broker.go
+++ b/pkg/reconciler/broker/broker.go
@@ -242,6 +242,7 @@ func (r *Reconciler) reconcileDLXDispatcherDeployment(ctx context.Context, b *ev
 			Broker: b,
 			Image:  r.dispatcherImage,
 			//ServiceAccountName string
+			Delivery:           b.Spec.Delivery,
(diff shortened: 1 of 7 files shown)


pkg/reconciler/broker/broker.go
```diff
@@ -242,6 +242,7 @@ func (r *Reconciler) reconcileDLXDispatcherDeployment(ctx context.Context, b *ev
 			Broker: b,
 			Image:  r.dispatcherImage,
 			//ServiceAccountName string
```
gabo1208, line 244: nit the comment goes in the line below

gabo1208 (commented):
Waiting for test to pass to approve

gabo1208 (approved):

PR #726, opened 2022-04-21
Adds upgrade and uninstall instructions to broker and source.

### LINKED ISSUE #413: Audit and document install & upgrade instructions for eventing-rabbitmq and its dependencies
Document for[ Operator Based Broker](https://github.com/knative-sandbox/eventing-rabbitmq/blob/main/broker/README.md#rabbitmq-operator-based-broker) in the repo

See #450  for details for RabbitMQ Operators
See https://knative.dev/docs/install/upgrade/upgrade-installation/ for Knative Eventing Upgrade Process

Document
- [ ] Install
- [ ] Upgrade
- [ ] Uninstall
- [ ] Add reference to [Production RMQ Setup](https://github.com/rabbitmq/cluster-operator/blob/a7fb232bf78c5ce735294ccc49da793a6d585ee9/docs/examples/production-ready/README.md#production-example)

gab-satchi:
/kind documenation

Fixes #413 

One thing that stood out while writing the docs, the recommendation for installing the operators is a very loose "install the latest". The cluster operator doesn't follow a strict [semver](https://github.com/rabbitmq/cluster-operator/blob/main/version_guidelines.md) and expects anyone using it to read through release notes to highlight any potential impact. This repo uses specific versions of both operators during testing and we should be prescriptive of which versions users should install for a given version of `eventing-rabbitmq`. It may make sense to include the recommended versions of the operators as a part of our release notes so it doesn't get outdated in docs and users can expect a consistent experience.

### DIFF
--- a/docs/broker/README.md
+++ b/docs/broker/README.md
@@ -7,11 +7,13 @@ RabbitMQ *is a Messaging Broker* - an intermediary for messaging. It gives your 
 # Table of Contents
 
 - [Installation](#installation)
-- [Autoscaling](#autoscaling-optional)
 - [Customizations](#customizations)
-- [Additional Resources](#additional-resources)
 - [Delivery Failures and Delivery Spec](#delivery-failures-and-delivery-spec)
 - [Troubleshooting](#troubleshooting)
+- [Next Steps](#next-steps)
+- [Additional Resources](#additional-resources)
+- [Upgrade](#upgrade)
+- [Uninstall](#uninstall)
 
 ## Installation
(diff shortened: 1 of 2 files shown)


ChunyiLyu (commented):
Looks good. One small suggestion: do you think it's worth mentioning that the upgrade step will cover just the rabbitmq-eventing components and rabbitmq operators, and rabbitmqcluster itself won't be upgraded as part of the step?

ChunyiLyu (approved):
LGTM

PR #771, opened 2022-05-17
Deployments resources

### LINKED ISSUE #703: Add `resource.[requests|limits]` to reconciled objects
**Problem**
Skimming through the `reconciler` codebase/folders, it seems that none of the deployments/pods created have any `resources` specified. This can lead to performance and stability issues, especially not meeting the documented scaling behavior (#408).

Operators and users.

**Exit Criteria**
Documented scaling limits (benchmarks) are met even when deploying the resources in a constrained (overutilized) cluster.

**Time Estimate (optional):**
Implementation is straightforward (add `requests` and/or `limits`), but finding the right values might take time (in benchmarks).

Even without benchmarking, adding sane defaults (at least `requests`) should be done for the various reconciled resources.

cc/ @salaboy @gabo1208

gabo1208:
# Changes

- 🎁 Now the Broker's Ingress and Dispatcher have deployment resource requests and limits
- 🎁 Now the Source's Receive Adapter have deployment resources requests and limits
Both are based on an overall stimation with a scenario where 1500msgs/s are sent over a parallelism of 1000 

/kind enhancement

Fixes #703 

**Release Note**

```release-note
- 🎁 Now the Broker's Ingress and Dispatcher have deployment resource requests and limits
- 🎁 Now the Source's Receive Adapter have deployment resources requests and limits
```

### DIFF
--- a/.github/workflows/kind-conformance.yaml
+++ b/.github/workflows/kind-conformance.yaml
@@ -24,7 +24,7 @@ jobs:
 
         # Map between K8s and KinD versions.
         # This is attempting to make it a bit clearer what's being tested.
-        # See: https://github.com/kubernetes-sigs/kind/releases/tag/v0.11.1
+        # See: https://github.com/kubernetes-sigs/kind/releases/tag/v0.12.0
         include:
         - k8s-version: v1.22.1
           kind-image-sha: sha256:100b3558428386d1372591f8d62add85b900538d94db8e455b66ebaf05a3ca3a
--- a/.github/workflows/kind-e2e.yaml
+++ b/.github/workflows/kind-e2e.yaml
@@ -18,12 +18,12 @@ jobs:
         - v1.22.1
         - v1.23.3
         knative-version:
-        - 1.0.2
-        - 1.1.0
+        - 1.3.2
+        - 1.4.0
 
         # Map between K8s and KinD versions.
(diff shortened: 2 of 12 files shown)


.github/workflows/kind-conformance.yaml
```diff
@@ -24,7 +24,7 @@ jobs:
 
         # Map between K8s and KinD versions.
         # This is attempting to make it a bit clearer what's being tested.
-        # See: https://github.com/kubernetes-sigs/kind/releases/tag/v0.11.1
+        # See: https://github.com/kubernetes-sigs/kind/releases/tag/v0.12.0
```
gab-satchi, line 27: We should probably update the knative version used for conformance tests. I see e2e is getting updated, conformance tests likely still using 1.1.0 (from Makefile)

gab-satchi (commented):
I haven't looked into the failures carefully but both the conformance test runs timed out after an hour. That may be an indicator that the settings are too aggressively low.

gabo1208:
> I haven't looked into the failures carefully but both the conformance test runs timed out after an hour. That may be an indicator that the settings are too aggressively low.

I think is the other case, cause with one extra kind node the tests run. So maybe i should set even lower the requests

gab-satchi (approved):
/lgtm

PR #800, opened 2022-06-03
Source simplification

### LINKED ISSUE #765: Redesigning RabbitMQSource API
**Problem**
There are a few things that stand out with the current RabbitMQSource API:
- When using RabbitMQ Cluster, there's no easy way to simply reference it in RabbitMQSource and have it connect. The user has to edit their `default-user` secret, add the uri and reference the secret in the RabbitMQSource.
- Some options like Exchange type don't work
- There are options exposed through the various resource configs (Exchange, Queue) that don't really add value or are a no-op to the Source's use case (more in context below)

Which persona is this feature for?
System Integrator

**Exit Criteria**
Provided a redesign is warranted, the following needs to be done
Finalize API and then fill out a task list
- [x] Limit the actual use cases for the current Source to avoid broken ones #795 #800 
- [x] Start deprecation notices on the Source Readme #795 
- [ ] Send email to the community communicating the Source API change
- [ ] Finalize the change discussion and definition for the Source 1.0
- [x] Trim all the deprecated Source Spec parameters and implement the new validations #800 
- [x] Use RabbitMQ's Cluster reference in the source #801
- [ ] Update docs and samples
- [ ] Upgrade instructions or conversion webhooks for sources

**Time Estimate (optional):**

The following is the current spec along with comments highlighting areas for discussion:

``` yaml
kind: RabbitmqSource
metadata:
  name: rabbitmq-source
spec:
  broker: "rabbitmq:5672/" # RabbitMQBroker uri. Currently used by the adapter deployment, can be simplified into RabbitMQClusterReference
  user: # Only used by the adapter deployment, can be simplified into RabbitMQClusterReference
    secretKeyRef:
      name: rabbitmq-default-user
      key: username
  password: # Only used by the adapter deployment, can be simplified into RabbitMQClusterReference
    secretKeyRef:
      name: rabbitmq-default-user
      key: password
  connectionSecret:
    name: "some-secret"
  channelConfig:
    parallelism: 1
    globalQos: false # The global QoS won't have any effect. The adapter uses a single channel off of a connection. Can just hardcode to false and not expose as an option
  exchangeConfig:
    name: "eventing-rabbitmq-source-exchange"
    type: "fanout" # The type (and following attributes) only matter when it isn't predeclared. It's a 1:1 setup of Exchange and Queue when predeclared is false. No need to have type configurable
    durable: true # Should hardcode to true and not be configurable
    autoDelete: false # Should hardcode to false and not be configurable. If exchange gets deleted, it would likely just get recreated during the next reconcile
  queueConfig:
    name: "eventing-rabbitmq-source-queue"
    routingKey: "some-key" # Routing key (and following attributes) only matter when it isn't predeclared. Routing key not needed in a 1:1 setup of Exchange + Queue
    durable: true # Should hardcode to true and not be configurable
    autoDelete: false # Should hardcode to false and not be configurable.
  predeclared: false
  retry: 5
  backoffPolicy: "linear"
  backoffDelay: "5ms"
  serviceAccountName: "some-account-name"
  vhost: "/"
  sink:
    ref:
      apiVersion: serving.knative.dev/v1
      kind: Service
      name: rabbitmq-source-sink
      namespace: source-demo

```

The following is a suggestion for consolidating all the fields required for connecting to a RabbitMQ server into a block in the spec called `RabbitmqClusterReference`. Copied from what's used in the Topology Operator:

``` golang
type RabbitmqClusterReference struct {
	// The name of the RabbitMQ cluster to reference.
	// Have to set either name or connectionSecret, but not both.
	// +kubebuilder:validation:Optional
	Name string `json:"name,omitempty"`
	// The namespace of the RabbitMQ cluster to reference.
	// Defaults to the namespace of the requested resource if omitted.
	// +kubebuilder:validation:Optional
	Namespace string `json:"namespace,omitempty"`
	// Secret contains the http management uri for the RabbitMQ cluster.
	// The Secret must contain the key `uri`, `username` and `password` or operator will error.
	// Have to set either name or connectionSecret, but not both.
	// +kubebuilder:validation:Optional
	ConnectionSecret *corev1.LocalObjectReference `json:"connectionSecret,omitempty"`
}
```

If all of the suggestions from above are applied (TODO: update after discussions), the new RabbitMQ Source spec would be:
``` yaml
kind: RabbitmqSource
metadata:
  name: rabbitmq-source
spec:
  rabbitmqClusterReference:
    name: "rmqCluster"
    namespace: "default"
  parallelism: 1 # Channel config flattened
  rmqResources:
    exchangeName: "eventing-rabbitmq-source-exchange" # Can also be implied if unset while predeclared is false
    queueName: "eventing-rabbitmq-source-queue" # Can also be implied if unset while predeclared is false
    predeclared: false
  delivery: # introduce deliverySpec similar to Broker/Trigger
    retry: 5
    backoffPolicy: "linear"
    backoffDelay: "5ms"
  serviceAccountName: "some-account-name"
  vhost: "/"
  sink:
    ref:
      apiVersion: serving.knative.dev/v1
      kind: Service
      name: rabbitmq-source-sink
      namespace: source-demo
```

If it's a predeclared setup, the spec get a bit simpler where exchangeName can become optional. Following is an example with predeclared non-managed RabbitMQ server:

```yaml
kind: RabbitmqSource
metadata:
  name: rabbitmq-source
spec:
  rabbitmqClusterReference:
    connectionSecret:
      name: "some-secret-name" # namespace is inferred as it's a localObjectReference
  parallelism: 1 # Channel config flattened
  rmqResources:
    exchangeName: "" # optional. won't get used when predeclared
    queueName: "eventing-rabbitmq-source-queue" # required as the dispatcher will consume from this queue
    predeclared: true
  delivery: # introduce deliverySpec similar to Broker/Trigger  
    retry: 5
    backoffPolicy: "linear"
    backoffDelay: "5ms"
  serviceAccountName: "some-account-name"
  vhost: "/"
  sink:
    ref:
      apiVersion: serving.knative.dev/v1
      kind: Service
      name: rabbitmq-source-sink
      namespace: source-demo
```

gabo1208:
# Changes

- 🎁 removed deprecated params from the new source spec

/kind api-change deprecation

Fixes #765

**Release Note**

```release-note
Now the Source has been simplified with less config and use cases, so it should be cleaner to setup and get it up and running
```


config/source/300-rabbitmqsource.yaml
```diff
@@ -81,19 +81,10 @@ spec:
               exchangeConfig:
```
gab-satchi, line 81: Is the flattening of the exchange and queue configs into `rmqResources` going to a part of this PR as well?

config/source/300-rabbitmqsource.yaml
```diff
@@ -126,18 +117,9 @@ spec:
...
                   name:
                     description: Name of the queue to bind to; required value.
                     type: string
-                  routingKey:
-                    description: Routing key of the messages to be received. Multiple
-                      routing keys can be specified separated by commas. e.g. key1,key2
-                    type: string
                 required:
                 - name
                 type: object
```
gab-satchi, line 125: Plan was to move retry and other deliverySpec options into a `delivery` struct similar to the Broker.

gab-satchi (commented):
Couple of things from the proposed source spec is missing. Not sure if that was going to be tackled in a follow-up PR.

config/source/300-rabbitmqsource.yaml
gabo1208: I wrote about that in the issue, lets discuss it there

config/source/300-rabbitmqsource.yaml
gabo1208: Yep that sounds good, but let's do it on another PR if that is ok

docs/source/README.md
```diff
@@ -288,35 +263,11 @@ The Source will provide output information about readiness or errors via the
...
-    These exchanges are useful for robust temporary topologies or when you require
-    binding durable queues to auto-deleted exchanges.
-
-    7. Note: RabbitMQ declares the default exchange types like 'amq.fanout' as
+    2. Note: RabbitMQ declares the default exchange types like 'amq.fanout' as
     durable, so queues that bind to these pre-declared exchanges must also be
     durable.
 
-    8. Optional amqp.Table of arguments that are specific to the server's implementation of
+    3. Optional amqp.Table of arguments that are specific to the server's implementation of
```
gab-satchi, line 270: Is this relevant still? I don't think there's a field we have for this

pkg/rabbit/exchange.go
```diff
@@ -58,9 +58,6 @@ func NewExchange(args *ExchangeArgs) *rabbitv1beta1.Exchange {
 		exchangeName = args.Name
 	} else if args.Source != nil {
 		ownerReference = *kmeta.NewControllerRef(args.Source)
-		durable = args.Source.Spec.ExchangeConfig.Durable
```
gab-satchi, line 61: These 3 can stop being variables now and just hardcoded below.

gab-satchi, line 61: These 3 can stop being variables now and just hardcoded below.

pkg/rabbit/queue.go
```diff
@@ -52,8 +52,6 @@ func NewQueue(args *QueueArgs) *rabbitv1beta1.Queue {
 
 	// queue configurations for source
 	if args.Source != nil {
-		durable = args.Source.Spec.QueueConfig.Durable
-		autoDelete = args.Source.Spec.QueueConfig.AutoDelete
```
gab-satchi, line 56: same as above commend. no longer needed to be vars

gab-satchi, line 56: same as above commend. no longer needed to be vars

pkg/utils/helpers_test.go
```diff
@@ -0,0 +1,55 @@
...
+			expectedPolicy: eventingduckv1.BackoffPolicyExponential,
+		},
+		"should return exponential policy when policy string is exponential": {
+			policyString:   "exponential",
+			expectedPolicy: eventingduckv1.BackoffPolicyExponential,
+		},
+		"should return linear policy when policy string is linear": {
+			policyString:   "linear",
+			expectedPolicy: eventingduckv1.BackoffPolicyLinear,
+		},
```
gab-satchi, line 42: if you add another case for unknown type, then codecov would stop complaining.

gab-satchi, line 42: if you add another case for unknown type, then codecov would stop complaining.

docs/source/README.md
gabo1208: true, removed

gab-satchi (approved):
/lgtm

PR #801, opened 2022-06-07
Source cluster reference

gabo1208:
# Changes

- 🎁 Now the Source uses a RabbitCluster Reference like the Broker
- 🧹 Refactored some helper methods regarding rabbitmq url extraction
- 🧹 Adapted tests, samples and docs

/kind api-change deprecation

Fixes #

**Release Note**

```release-note
 RabbitMQ Source changed to use a RabbitCluster Reference like the Broker
```


gab-satchi (commented):
just did a first pass

gab-satchi:
nit with the release notes: Let's not start all points with "Now", that's kind of implied. It can just be something like `Rabbitmq sources changed to use rabbitmq cluster references for connectivity"

pkg/apis/eventing/v1alpha1/rabbitmq_validation.go
```diff
@@ -46,22 +46,19 @@ func (r *RabbitmqBrokerConfig) Validate(ctx context.Context) *apis.FieldError {
...
 	} else {
 		if r.Spec.RabbitmqClusterReference.Name == "" {
-			errs = errs.Also(apis.ErrMissingField("name").ViaField("rabbitmqClusterReference").ViaField("spec"))
-		}
-		if r.Spec.RabbitmqClusterReference.Namespace == "" {
-			errs = errs.Also(apis.ErrMissingField("namespace").ViaField("rabbitmqClusterReference").ViaField("spec"))
-		}
-		if r.Spec.RabbitmqClusterReference.ConnectionSecret != nil {
+			if r.Spec.RabbitmqClusterReference.ConnectionSecret == nil {
+				return apis.ErrMissingField("name").ViaField("rabbitmqClusterReference").ViaField("spec")
```
gab-satchi, line 54: We can't make this change right now. RabbitmqBrokerConfig does not accept a connection secret in its current state. It'll be a new feature to add that capability.

pkg/apis/sources/v1alpha1/rabbitmq_validation.go
```diff
@@ -33,6 +33,18 @@ func (current *RabbitmqSource) Validate(ctx context.Context) *apis.FieldError {
 		}
 	}
 
+	if current.Spec.RabbitmqClusterReference == nil {
+		return apis.ErrMissingField("rabbitmqClusterReference").ViaField("spec")
+	} else {
+		if current.Spec.RabbitmqClusterReference.Name == "" {
```
gab-satchi, line 39: With this, can namespace be omitted? Where are we handling the defaulting for that?

gabo1208: From the ClusterReference code:
```
type RabbitmqClusterReference struct {
	// The name of the RabbitMQ cluster to reference.
	// Have to set either name or connectionSecret, but not both.
	// +kubebuilder:validation:Optional
	Name string `json:"name,omitempty"`
	// The namespace of the RabbitMQ cluster to reference.
	// Defaults to the namespace of the requested resource if omitted. <-- Here
	// +kubebuilder:validation:Optional
	Namespace string `json:"namespace,omitempty"`
	// Secret contains the http management uri for the RabbitMQ cluster.
	// The Secret must contain the key `uri`, `username` and `password` or operator will error.
	// Have to set either name or connectionSecret, but not both.
	// +kubebuilder:validation:Optional
	ConnectionSecret *corev1.LocalObjectReference `json:"connectionSecret,omitempty"`
}
```

pkg/apis/eventing/v1alpha1/rabbitmq_validation.go
gabo1208: Ok ok. I'll roll it back, i wanted to do a common method to validate both anyway :p! Ty

pkg/apis/sources/v1alpha1/rabbitmq_validation.go
gab-satchi: For context, with the RabbitmqBrokerConfig, I made both name and namespace required fields

pkg/apis/eventing/v1alpha1/rabbitmq_validation_test.go
```diff
@@ -78,17 +78,6 @@ func TestValidateRabbitmqBrokerConfig(t *testing.T) {
 			wantErr:  true,
 			isUpdate: false,
 		},
-		"missing rabbitmqClusterReference.namespace": {
```
gab-satchi, line 81: we can revert this change as well since the source code changes have been reverted for this file it looks like.

pkg/apis/sources/v1alpha1/rabbitmq_validation.go
```diff
@@ -33,6 +33,21 @@ func (current *RabbitmqSource) Validate(ctx context.Context) *apis.FieldError {
...
+			return errs.Also(apis.ErrMissingField("name").ViaField("rabbitmqClusterReference").ViaField("spec"))
+		}
+		if current.Spec.RabbitmqClusterReference.Namespace == "" {
+			return errs.Also(apis.ErrMissingField("namespace").ViaField("rabbitmqClusterReference").ViaField("spec"))
+		}
+		if current.Spec.RabbitmqClusterReference.ConnectionSecret != nil {
+			return apis.ErrDisallowedFields("connectionSecret").ViaField("rabbitmqClusterReference").ViaField("spec")
```
gab-satchi, line 47: Might be copy pasta issue with the RabbitmqBrokerConfig, but for sources `connectionSecret` is an allowed field

pkg/apis/sources/v1alpha1/rabbitmq_validation_test.go
```diff
@@ -291,3 +303,64 @@ func TestRabbitmqSourceExchangeConfig(t *testing.T) {
...
+			},
+			wantErr: true,
+		},
+		"just connection secret": {
+			spec: &RabbitmqSourceSpec{
+				Predeclared: true,
+				RabbitmqClusterReference: &v1beta1.RabbitmqClusterReference{
+					Namespace: "test", ConnectionSecret: &v1.LocalObjectReference{Name: "test"}},
+			},
+			wantErr: true,
```
gab-satchi, line 348: this should be allowed

gabo1208: No, since there are a lot of changes here, the connection secret relies on the broker, username and password params that are been removed, on the next PR that is been solved for the samples to work. that's why I'm modifying these tests too

pkg/apis/sources/v1alpha1/rabbitmq_validation.go
gabo1208: answered on the other comment

PR #810, opened 2022-06-10
Source params reordering with docs and samples updates

gabo1208:
# Changes

- 🎁 Source's Spec RabbitMQ related parameters have been grouped in the RabbitMQResourcesConfig Spec
- 🎁 Source's Spec Delivery related parameters have been grouped in the Delivery Spec
- 🎁 Added support for ClusterReference ConnectionSecret option to the Source
- 🧹 Updating Source's docs and samples to reflect this changes

/kind api-change documentation

Fixes #

**Release Note**

```release-note
- BREAKING: 🎁 Source's Spec RabbitMQ related parameters have been grouped in the RabbitMQResourcesConfig Spec, Source's Spec Delivery related parameters have been grouped in the Delivery Spec
- 🎁 Added support for ClusterReference ConnectionSecret option to the Source
```


pkg/adapter/adapter.go
```diff
@@ -41,30 +41,18 @@ import (
...
-	ExchangeConfig ExchangeConfig
-	QueueConfig    QueueConfig
+	RabbitURL     string        `envconfig:"RABBIT_URL" required:"true"`
+	Vhost         string        `envconfig:"RABBITMQ_VHOST" required:"false"`
+	Predeclared   bool          `envconfig:"RABBITMQ_PREDECLARED" required:"false"`
+	Retry         int           `envconfig:"HTTP_SENDER_RETRY" required:"false"`
+	BackoffPolicy string        `envconfig:"HTTP_SENDER_BACKOFF_POLICY" required:"false"`
+	BackoffDelay  time.Duration `envconfig:"HTTP_SENDER_BACKOFF_DELAY" default:"50ms" required:"false"`
+	Parallelism   int           `envconfig:"RABBITMQ_CHANNEL_CONFIG_PARALLELISM" default:"1" required:"false"`
+	ExchangeName  string        `envconfig:"RABBITMQ_EXCHANGE_CONFIG_NAME" required:"false"`
```
gab-satchi, line 54: nit: the env vars can be simplified. No need for `CONFIG` in there. Can be `RABBITMQ_EXCHANGE_NAME` for ex

pkg/apis/sources/v1alpha1/rabbitmq_validation.go
```diff
@@ -19,40 +19,37 @@ package v1alpha1
...
-		}
-		if current.Spec.RabbitmqClusterReference.Namespace == "" {
-			return errs.Also(apis.ErrMissingField("namespace").ViaField("rabbitmqClusterReference").ViaField("spec"))
-		}
-		if current.Spec.RabbitmqClusterReference.ConnectionSecret != nil {
+			if current.Spec.RabbitmqClusterReference.ConnectionSecret == nil {
+				return apis.ErrMissingField("name").ViaField("rabbitmqClusterReference").ViaField("spec")
+			}
+		} else if current.Spec.RabbitmqClusterReference.ConnectionSecret != nil {
 			return apis.ErrDisallowedFields("connectionSecret").ViaField("rabbitmqClusterReference").ViaField("spec")
```
gab-satchi, line 46: It'll be worthwhile having a more descriptive error here rather than `connectionSecret` is `Disallowed`. Something like both name+namespace and connection secret cannot be set.

pkg/rabbit/service.go
```diff
@@ -203,7 +204,35 @@ func isReady(conditions []v1beta1.Condition) bool {
 	return numConditions == 0
 }
 
-func (r *Rabbit) RabbitMQURL(ctx context.Context, clusterRef *rabbitv1beta1.RabbitmqClusterReference) (*url.URL, error) {
+func (r *Rabbit) RabbitMQURL(ctx context.Context, clusterRef *rabbitv1beta1.RabbitmqClusterReference, parentNamespace string) (*url.URL, error) {
+	if clusterRef.ConnectionSecret != nil {
+		ns := "default"
+		if clusterRef.Namespace != "" {
+			ns = clusterRef.Namespace
+		} else if parentNamespace != "" {
```
gab-satchi, line 212: parentNamespace seems to be broker.namespace. Would that ever be empty? 

I get why we're defaulting here (because the rabbitmqClusterReference struct has that description), but I'm wondering if this would be intuitive enough for a user. Having it default to the Broker's namespace would need to be documented and there's no guarantee a user would actually see it. I think having namespace be required or just having a simple `+kubebuilder:default:=default` if we want to leave it optional would be better.

pkg/apis/sources/v1alpha1/rabbitmq_validation.go
gabo1208: Ok, but since there is a refactor comming regarding this validation, can this be moved to the next pr?

pkg/apis/sources/v1alpha1/rabbitmq_validation.go
```diff
@@ -19,40 +19,37 @@ package v1alpha1
...
 		}
 	}
 
 	if apis.IsInUpdate(ctx) {
```
gab-satchi, line 50: We'd want this block to be the first block. So if the user goes from having the spec populated to being empty for example, we'd want to error and say spec is immutable. Right now, the webhook would complain about `RabbitmqResourcesConfig` is missing which can mislead the user into thinking the spec is mutable.

gab-satchi:
another nit regarding the release notes: We can just have a single line or a few lines regarding the Source API changing. With a `BREAKING:` prefix. With the release notes in the last few PRs, they highlight individual things the PRs are doing and that's too detailed for release notes.

pkg/apis/eventing/v1alpha1/rabbitmq_defaults.go
```diff
@@ -18,4 +18,12 @@ package v1alpha1
 
 import "context"
 
-func (r *RabbitmqBrokerConfig) SetDefaults(ctx context.Context) {}
+func (r *RabbitmqBrokerConfig) SetDefaults(ctx context.Context) {
+	if r.Namespace == "" {
```
gab-satchi, line 22: isn't this the k8s resource namespace? we shouldn't be modifying that

pkg/apis/eventing/v1alpha1/rabbitmq_defaults.go
```diff
@@ -18,4 +18,12 @@ package v1alpha1
...
+		r.Namespace = "default"
+	}
+
+	if r.Spec.RabbitmqClusterReference != nil && r.Spec.RabbitmqClusterReference.Namespace == "" {
+		r.Spec.RabbitmqClusterReference.Namespace = r.Namespace
```
gab-satchi, line 27: this should just be `r.Spec.RabbitmqClusterReference.Namespac = default`

pkg/apis/sources/v1alpha1/rabbitmq_defaults.go
```diff
@@ -18,4 +18,12 @@ package v1alpha1
 
 import "context"
 
-func (r *RabbitmqSource) SetDefaults(ctx context.Context) {}
+func (r *RabbitmqSource) SetDefaults(ctx context.Context) {
+	if r.Namespace == "" {
+		r.Namespace = "default"
```
gab-satchi, line 23: same as above

pkg/apis/sources/v1alpha1/rabbitmq_validation.go
gab-satchi: just reiterating that the update bit should be the first thing we check for.

gab-satchi (commented):
some minor things

gab-satchi (approved):
/lgtm
/hold for tests

PR #814, opened 2022-06-13
Fix DLQ flake

### LINKED ISSUE #792: Fix DeliverySpec flake
**Describe the bug**
From time to time we see flakes like this: https://github.com/knative-sandbox/eventing-rabbitmq/runs/6584787394?check_suite_focus=true#step:7:69846

 **Expected behavior**
Should be always green

 **To Reproduce**
[Steps to reproduce the behavior.](https://github.com/gab-satchi/eventing-rabbitmq/blob/observe-flake/run-conformance.sh)

 **Knative release version**

 **Additional context**
This is a confirmed problem within RabbitMQ itself. While fix is created and goes through all release cycles the workaround is to serialize queue and policy creation.

ikvmw:
- Separates queue and policy reconciliation to prevent a race condition on RMQ server.

/kind bug

Fixes #792

### DIFF
--- a/pkg/rabbit/service.go
+++ b/pkg/rabbit/service.go
@@ -89,29 +89,32 @@ func (r *Rabbit) ReconcileQueue(ctx context.Context, args *QueueArgs) (Result, e
 		return Result{}, err
 	}
 
-	policyReady := true
-	if args.DLXName != nil {
-		want := NewPolicy(args)
-		policy, err := r.RabbitmqV1beta1().Policies(args.Namespace).Get(ctx, queue.Name, metav1.GetOptions{})
-		if apierrs.IsNotFound(err) {
-			logging.FromContext(ctx).Debugw("Creating rabbitmq policy", zap.String("name", queue.Name))
-			policy, err = r.RabbitmqV1beta1().Policies(args.Namespace).Create(ctx, want, metav1.CreateOptions{})
-		} else if err != nil {
-			return Result{}, fmt.Errorf("error creating queue policy: %w", err)
-		} else if !equality.Semantic.DeepDerivative(want.Spec, policy.Spec) {
-			desired := policy.DeepCopy()
-			desired.Spec = want.Spec
-			logging.FromContext(ctx).Debugw("Updating rabbitmq policy", zap.String("name", queue.Name))
-			policy, err = r.RabbitmqV1beta1().Policies(args.Namespace).Update(ctx, desired, metav1.UpdateOptions{})
-		}
-		if err != nil {
-			return Result{}, err
-		}
-		policyReady = isReady(policy.Status.Conditions)
+	return Result{
+		Name:  want.Name,
+		Ready: isReady(queue.Status.Conditions),
+	}, nil
+}
+
+func (r *Rabbit) ReconcileDLQPolicy(ctx context.Context, args *QueueArgs) (Result, error) {
+	want := NewPolicy(args)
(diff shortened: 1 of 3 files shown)


test/conformance/resources/rabbitmqcluster/rabbitmqcluster.yaml
```diff
@@ -12,3 +12,13 @@ spec:
...
       memory: "1Gi"
+  override:
+    statefulSet:
+      spec:
+        template:
+          spec:
+            containers:
+            - name: rabbitmq
+              env:
+              - name: ERL_MAX_PORTS
```
gab-satchi, line 23: Is this change still needed?

gab-satchi (commented):
Changes lgtm. just a question about overriding the ERL_MAX_PORTS

test/conformance/resources/rabbitmqcluster/rabbitmqcluster.yaml
```diff
@@ -12,3 +12,13 @@ spec:
...
+                value: "4096"
```
gabo1208, line 24: why this limit?

gabo1208 (approved):

gabo1208:
Just a comment but seems good to me

test/conformance/resources/rabbitmqcluster/rabbitmqcluster.yaml
ikvmw: mentioned it today during retro - https://github.com/rabbitmq/cluster-operator/issues/959

gab-satchi: But we're not experiencing a crash loop with the node not starting though right? Our RMQ version is also newer (3.10) I believe so maybe that's why we're bypassing the issue?

ikvmw: it doesn't depend on RMQ version. it's a combination of host setup and memory limit. When Erlang starts it tries to allocate memory stuctures for all available FDs. For example:

```
cat /proc/sys/fs/file-max 
9223372036854775807
```

So the fact that it works locally and on GH is a pure luck

gab-satchi: I can just barely grok what's happening with Erlang here. I understand that with a too high FD limit, Erlang can cause RMQ broker to OOM as it tries to allocate something per FD. 

But it does look like this change isn't directly tied to the DLQ flake we're seeing in conformance tests. Maybe we can omit this change for this PR to unblock it and then open something separate to discuss this? We also have RMQClusters defined elsewhere in docs and setup instructions so we'll likely need a broader change than just the test clusters.

ikvmw: >> I can just barely grok what's happening with Erlang here. I understand that with a too high FD limit, Erlang can cause RMQ broker to OOM as it tries to allocate something per FD.

you got it completely right. we have to limit the subset erlang sees because we don't control test environments. it's like restricting max file size while doing uploads handler, etc. 

Having really high fd limit is not a problem of course. The problem is this particular software pattern of preallocating stuff.

It's of cpurse possible for me to have this change in a separate PR. However, on my system tests can't be run without capping ERL_MAX_PRTS. So that's why it goes as a package.

PR #825, opened 2022-06-19
Make sure rabbitmq set up properly and test env allows reproducibility

ikvmw:
See https://github.com/rabbitmq/cluster-operator/issues/959

Unfortunately docker doesn't give enough isolation/reproducability and kernel's fs.max-fds makes into containers. Which means our tests/samples can't run on arbitrary systems - i.e. test env not sealed. Please take a look inside linked operator issue.

**TODO:** Let's decide what are the good values for samples and performance cluster.

cc @coro

# Changes

- :bug: Limit ERL_MAX_PORTS explicitly, since host's fs.fd-max leaks into containers. 

/kind bug

**Release Note**

```release-note
Limit ERL_MAX_PORTS explicitly in tests, since host's fs.fd-max leaks into containers. the values provided in samples might be too concervative, please adjust to the taste. 
```

### DIFF
--- a/samples/broker-trigger/quick-setup/200-rabbitmq.yaml
+++ b/samples/broker-trigger/quick-setup/200-rabbitmq.yaml
@@ -6,3 +6,13 @@ metadata:
   namespace: broker-trigger-demo
 spec:
   replicas: 1
+  override:
+    statefulSet:
+      spec:
+        template:
+          spec:
+            containers:
+            - name: rabbitmq
+              env:
+              - name: ERL_MAX_PORTS
+                value: "4096"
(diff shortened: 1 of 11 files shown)


coro:
This change looks correct from the operator's point of view, and is the workaround that we recommend in https://github.com/rabbitmq/cluster-operator/issues/959. We can't prevent the host from leaking into the RabbitMQ Pods on an operator level, so it's down to the user to limit the max ports on their system via the very widely-used override interface.

ikvmw:
Added more. Decided to keep performance tests as is, since 1) they work 2) it's hard to predict what makes sense here, since folks likely would want to play with concurrency parameter.

PR #827, opened 2022-06-19
Override and escape component provided by controller.

### LINKED ISSUE #826: Reconciler ConfigMap to Env hack mismanages ComponentName
**Describe the bug**
This reports started as Knative Slack thread: https://knative.slack.com/archives/C02N2AK2K1P/p1652037600027259
User reports that /metrics endpoint of our ingress returns nothing and many errors logged.
```
2022/06/14 17:12:02 Failed to export to Prometheus: "rabbitmq-broker-controller_go_frees" is not a valid metric name
```

 **Expected behavior**
/metrics works and renders all our metrics

 **To Reproduce**
Just spin up our stuff, forward port 9090, touch with `curl localhost:9090/metrics` see nothing, see erros in logs.

 **Knative release version**
1.5.0

 **Additional context**
First question when looking at the error: why "rabbitmq-broker-controller"?
In prometheus world it's common to have metrics namespace dubbed as Component name or somthing like that. All metrics prefixed with string. In our case we have two problems:
- wrong prefix - should be something about ingress
- dashes `-` are not escaped (usually they just get replaced with `_`)

**Wrong prefix**
So how come ingress uses `rabbitmq-broker-controller` as prefix? Well it turns out that it comes via ConfigWatcher we have to inject obeservability-related config from the respective ConfigMap (https://github.com/ikvmw/eventing-rabbitmq/blob/main/pkg/reconciler/broker/controller.go#L62 and https://github.com/ikvmw/eventing-rabbitmq/blob/main/pkg/reconciler/broker/controller.go#L115). Later this ComponentName used to cache MetricsConfig https://github.com/ikvmw/eventing-rabbitmq/blob/main/vendor/knative.dev/eventing/pkg/reconciler/source/config_watcher.go#L173. This MetricsConfig later injected to ingress env as `K_METRICS_CONFIG`: https://github.com/ikvmw/eventing-rabbitmq/blob/main/pkg/reconciler/broker/resources/ingress.go#L84 as json. 
```
K_METRICS_CONFIG:  {"Domain":"knative.dev/eventing","Component":"rabbitmq-broker-controller","PrometheusPort":0,"PrometheusHost":"","ConfigMap":{}}      
```
Later ingress loads MetricsConfig from the env: https://github.com/ikvmw/eventing-rabbitmq/blob/main/cmd/ingress/main.go#L54 and https://github.com/ikvmw/eventing-rabbitmq/blob/main/pkg/utils/envconfig.go#L69.

**No escapement**
<placeholder>

ikvmw:
# Changes

- :bug: Fix metrics component name for user-facing components. Also escape it.

/kind bug

Fixes #826 

**Release Note**

```release-note
Fix metrics component name for user-facing components. Also escape it.
```

### DIFF
--- a/pkg/utils/envconfig.go
+++ b/pkg/utils/envconfig.go
@@ -3,6 +3,7 @@ package utils
 import (
 	"context"
 	"net/http"
+	"strings"
 
 	"go.uber.org/zap"
 	"knative.dev/pkg/logging"
@@ -72,7 +73,9 @@ func (e *EnvConfig) SetupMetrics(ctx context.Context) error {
 	if err != nil {
 		return err
 	}
+
 	if metricsConfig != nil {
+		metricsConfig.Component = strings.ReplaceAll(e.component, "-", "_")
 		if err := metrics.UpdateExporter(ctx, *metricsConfig, logger); err != nil {
 			return err
 		}


gab-satchi (approved):
/lgtm

PR #831, opened 2022-06-22
Source delivery fixes

### LINKED ISSUE #830: RabbitMQ source hangs when message cannot be processed and retry configuration is set
**Describe the bug**

Architecture:
`RabbitMQ` -> `RabbitMQ event source` -> `HTTP service that returns error`

Seems that when retry is configured, then exception occours
image

If retry values are not configured, then error is written in rabbitmq source log and next messages are processed

##### Example rabbitmq source 
```yaml
kind: RabbitmqSource
metadata:
  name: my-event-source
spec:
  broker: "rabbitmq:5672/"
  connectionSecret:
    name: rabbitmq
  user:
    secretKeyRef:
      name: rabbitmq
      key: username
  password:
    secretKeyRef:
      name: rabbitmq
      key: password
  exchangeConfig:
    type: "fanout"
    durable: true
    name: "my"
  queueConfig:
    name: "my"
    durable: true
    autoDelete: false
  channelConfig:
    parallelism: 1 
  predeclared: false
  sink:
    ref:
      apiVersion: v1
      kind: Service
      name: service-that-returns-error
  # Retry "forever" => 30 days" 
  retry: 2592000
  backoffPolicy: "linear"
  backoffDelay: "1s"
  
```
 **Expected behavior**
Message is retried

 **To Reproduce**
In sink section configure to service that returns error. Also configure retry policy (see example above)

 **Knative release version**
1.5.0

 **Additional context**
Also rabbitmq hangs up and source is not restarted, which results in broken rabbitmq eventing, so no messages can be delivered through

gabo1208:
# Changes

- 🐛 Fixed bug In the Source, when if the Retries Spec was set and an error response was sent by the sink, the system crashed because a miss configured backoff function

/kind  bug

Fixes #830

**Release Note**

```release-note
🐛 Fixed bug In the Source, when if the Retries Spec was set and an error response was sent by the sink, the system crashed because a miss configured backoff function
```

### DIFF
--- a/pkg/adapter/adapter.go
+++ b/pkg/adapter/adapter.go
@@ -22,7 +22,6 @@ import (
 	nethttp "net/http"
 	"strings"
 	"sync"
-	"time"
 
 	"github.com/NeowayLabs/wabbit"
 	"github.com/NeowayLabs/wabbit/amqp"
@@ -33,7 +32,7 @@ import (
 	"knative.dev/eventing-rabbitmq/pkg/rabbit"
 	"knative.dev/eventing-rabbitmq/pkg/utils"
 	"knative.dev/eventing/pkg/adapter/v2"
-	eventingduckv1 "knative.dev/eventing/pkg/apis/duck/v1"
+	v1 "knative.dev/eventing/pkg/apis/duck/v1"
 	"knative.dev/eventing/pkg/kncloudevents"
 	"knative.dev/eventing/pkg/metrics/source"
 	"knative.dev/pkg/logging"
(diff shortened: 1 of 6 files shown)


pkg/adapter/adapter.go
```diff
@@ -183,9 +186,21 @@ func (a *Adapter) ConsumeMessages(channel *wabbit.Channel,
 func (a *Adapter) PollForMessages(channel *wabbit.Channel,
 	queue *wabbit.Queue, stopCh <-chan struct{}) error {
 	logger := a.logger
+	var err error
+	if a.config.BackoffDelay != "" {
+		retriesInt32 = int32(a.config.Retry)
+		backoffPolicy := utils.SetBackoffPolicy(a.context, a.config.BackoffPolicy)
```
gab-satchi, line 192: `utils.SetBackoffPolicy` can still return a "". Should we continue checking for that?

gab-satchi (commented):
We should add/modify a test that exercises the retry capability of Source

PR #851, opened 2022-07-05
Samples and tests reorganization

gabo1208:
# Changes

- 🧹 Reorganized samples to be the same for both Broker and Source to we avoid duplicating code and README files
- 🧹 Reorganized e2e test to single files per CRD to avoid having lots of files with single test cases and no descriptive names

/kind documentation


gab-satchi (approved):
/lgtm

PR #856, opened 2022-07-08
Sets prefetch to a static 100. Update docs

### LINKED ISSUE #835: Default settings for parallelism and queue type causing performance degradation
**Describe the bug**
The default settings of a Broker/Trigger setup uses:
- Parallelism set to 1. This prefetches 1 message at a time and processes it using a single worker
- Queue type set to quorum. 

With quorum queues, the acks can take a while to process. The acks need to be written to disk and propagated to all the nodes in the cluster. This causes a bottle-neck where the consumer is "starving" when the consumer is fetching a single message at a time. 

Options:
- Set default queue type to `classic` since it's configurable and user can modify.
- Set a > 1 default for parallelism.
- Split parallelism into `prefetchCount` and `workers`.
  - option 1: prefetchCount isn't configurable but parallelism becomes number of workers. Prefetch is set to a static value like 100.
  - option 2: prefetchCount and workers can both be configurable. Can still default prefetch to be > 1 while allowing it to be configurable by users.

 **Knative release version**
upcoming release

 **Additional context**
The following are two runs of the same test. A producer producing 1000 messages a second and a single consumer consuming them. The key difference between the two is the queue type. We can see the drop in throughput when using Quorum queues with prefetch set to 1.

*Note:* This issue is only present with a very low prefetch number. The performance is pretty similar when running the same tests with parallelism set to 100.

**Classic Queue**
parallel-1-throughput

The throughput is around 500-600 until the producer finishes and then jumps to well over 1000.

**Quorum Queue**
parallel-1-throughput

The throughput barely breaks 400 even after the producer is finished.

gab-satchi:
# Changes
- Parallelism only affects number of workers. Prefetch set to an non-configurable value of 100

- 🗑️ parallelism no longer sets the prefetch count

/kind removal
/kind performance

Fixes #835 

```release-note
Parallelism only affects number of workers. Prefetch set to an non-configurable value of 100
```

/hold for reviews

### DIFF
--- a/cmd/dispatcher/main.go
+++ b/cmd/dispatcher/main.go
@@ -141,9 +141,9 @@ func (env *envConfig) setupRabbitMQ(ctx context.Context) {
 	}
 
 	err = env.channel.Qos(
-		env.Parallelism, // prefetch count
-		0,               // prefetch size
-		false,           // global
+		100,   // prefetch count
+		0,     // prefetch size
+		false, // global
 	)
 	if err != nil {
 		logging.FromContext(ctx).Fatal("Failed to create QoS: ", err)
--- a/docs/broker/README.md
+++ b/docs/broker/README.md
@@ -79,13 +79,9 @@ install
 [KEDA based autoscaler](https://github.com/knative-sandbox/eventing-autoscaler-keda).
 
 ### Trigger Parallelism Count
-Trigger has a configurable annotation `rabbitmq.eventing.knative.dev/parallelism`. The following are effects of setting this parameter to `n`:
-
-- Prefetch count is set to this value on the RabbitMQ channel and queue created for this trigger. The channel will receive a maximum of `n` number of messages at once.
-- The trigger will create `n` workers to consume messages off the queue and dispatch to the sink.
-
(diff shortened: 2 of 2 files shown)


cmd/dispatcher/main.go
```diff
@@ -141,9 +141,9 @@ func (env *envConfig) setupRabbitMQ(ctx context.Context) {
 	}
 
 	err = env.channel.Qos(
-		env.Parallelism, // prefetch count
-		0,               // prefetch size
-		false,           // global
+		100,   // prefetch count
```
gabo1208, line 144: Why not 1000? Are we sure that the 100 is a good value?

gab-satchi: Since it's a single consumer, a higher prefetch value would generally increase message delivery. On the flip side, memory usage will get impacted with a higher prefetch count. I chose 100 as it's something that's been tested in our performance tests. There isn't an ideal number that would work for all scenarios a user can come across.

gabo1208: Kk maybe for the future we could start playing with those numbers, but approving :)

gabo1208 (approved):

PR #868, opened 2022-07-20
Attempts to remove any leftover DLQ resources that are no longer needed

### LINKED ISSUE #465: Deleting trigger DLQ spec should clean up resources
Repro:
- Create a broker/trigger with a DLQ on the trigger spec
- Remove the DLQ from the trigger spec
- The binding between the exchange and the queue should be deleted
- The DLQ queue should be deleted
- The dispatcher on the queue should be deleted

gab-satchi:
# Changes
- 🐛  Passively tries to remove any leftover resources. Will ignore errors as it won't affect functionality.

/kind bug

Fixes #465 

```release-note
- Removes leftover RMQ resources if a Trigger's delivery spec is removed
```

### DIFF
--- a/pkg/rabbit/service.go
+++ b/pkg/rabbit/service.go
@@ -18,6 +18,7 @@ package rabbit
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"net/url"
 	"strings"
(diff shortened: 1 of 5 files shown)


gabo1208:
Thought that the topology operator was taking charge of this when setting the owner

gab-satchi:
Setting the owner will make sure the child objects get deleted when the parent is deleted. But it won't do anything when updating the parent object.

gabo1208 (approved):

Issue #566, opened 2022-01-10
Validate eventing-rabbitmq works with a RMQ Cluster configured with TLS


xtreme-sameer-vohra:
**Setup**
- Setup RMQ cluster configured with signed certificate
- Setup eventing-rabbitmq Broker that connects to the RMQ cluster using TLS address and can deliver messages to sink
- Setup eventing-rabbitmq Source that connects to the RMQ cluster using TLS address and can retrieve messages and deliver to sink

**[Persona:](https://github.com/knative/eventing/blob/main/docs/personas.md)**
Which persona is this feature for?
This feature enables Developers to comply with Administrator policies

**Exit Criteria**
The traffic between the following components is encrypted
- [ ] RMQ Cluster & Source
- [ ] RMQ Cluster & Broker + Trigger

**Time Estimate (optional):**

**Additional context (optional)**
https://github.com/knative/eventing/issues/5957
https://github.com/rabbitmq/messaging-topology-operator/issues/233

salaboy:
@ikvmw Is there something that we need to be aware of on this topic to connect to RabbitMQ with TLS? I believe that this can be done after MVP, but it will be great to know what exactly we should be doing here.. or if we expect to find issues.

gabo1208:
This week I've been learning an implementing TLS for the communication between the Eventing RabbitMQ and RabbitMQ using the topology operator. On our side there were some bits missing, so next week this should be ready to try.

xtreme-sameer-vohra:
Thanks @gabo1208 
What are the missing bits ?

gab-satchi:
/close

done by #942

PR #893, opened 2022-08-17
Enable tls

### LINKED ISSUE #818: RabbitMQ Port not configurable
**Problem**
Right now we don't have a way for a users to set a different port for RabbitMQ than the default 5672.

Users

**Exit Criteria**
- [x] Add a RabbitMQ port config option in the Broker
- [x] Add a RabbitMQ port config option in the Trigger
- [x] Add a RabbitMQ port config option in the Source

gabo1208:
# Changes

- 🎁 Enable Tls for RabbitMQ Broker for known CA authorities
- 🎁 Enable Tls for RabbitMQ Source for known CA authorities
- 🎁 Now the cluster port is configurable on the rabbitmq cluster secret on the `port` param (if you are not using the RabbitMQ Topology Operator)
- 🧹 Minor refactor to rabbit service helper

/kind enhancement

Fixes #818 #566

**Release Note**

```release-note
- 🎁 Enable Tls for RabbitMQ Broker for known CA authorities
- 🎁 Enable Tls for RabbitMQ Source for known CA authorities
- 🎁 RabbitMQ Cluster connection port is configurable on the rabbitmq cluster secret on the `port` param (if you are not using the RabbitMQ Topology Operator)
```

**Docs**


gab-satchi (commented):
Just did a quick pass. There are a few things that stand out:
- I don't think we should have a flag to say "ssl". When it's a RabbitMQCluster, we could get the tls config from that reference. So the ssl field is only useful for the `Secret` reference. It might be okay to do what messaging-topology-operator [does](https://github.com/rabbitmq/messaging-topology-operator/blob/58c2bdc604f7d4334b3e7e71f36effbbe7b5e6f5/rabbitmqclient/cluster_reference.go#L40) which is just rely on the uri to have an `http` or `https`.
- I see us switching the protocol based on the ssl flag, but not doing loading any of the system certs which may be required for self-sigend certs.
- self-signed certs process would need to be documented (can be punted to followup)
- This would be a feature to add some e2e tests for

gab-satchi:
If the self-signed bits aren't in then we should remove those from the release notes and the "Fixes". You'd need to add the ca certs when making the amqp call so it can be trusted.

config/broker/300-rabbitmqbrokerconfig.yaml
```diff
@@ -54,8 +54,8 @@ spec:
                   connectionSecret:
                     description: Secret contains the http management uri for the RabbitMQ
                       cluster. The Secret must contain the key `uri`, `username` and
-                      `password` or operator will error. Have to set either name or
-                      connectionSecret, but not both.
+                      `password` or operator will error, it may contain also `port` (default 5672).
```
gab-satchi, line 57: The Secret must contain the key `uri`, `username` and `password`. It may optionally contain `port` or will be defaulted to 5672

config/source/300-rabbitmqsource.yaml
```diff
@@ -77,8 +77,8 @@ spec:
                   connectionSecret:
                     description: Secret contains the http management uri for the RabbitMQ
                       cluster. The Secret must contain the key `uri`, `username` and
-                      `password` or operator will error. Have to set either name or
-                      connectionSecret, but not both.
+                      `password` or operator will error, it may contain also `port` (default 5672).
```
gab-satchi, line 80: The Secret must contain the key `uri`, `username` and `password`. It may optionally contain `port` or will be defaulted to 5672

docs/source/README.md
```diff
@@ -169,7 +169,7 @@ Source parameters
 |--------|--------|
 | `rabbitmqClusterReference` | A reference to a RabbitMQ Cluster |
 | `rabbitmqClusterReference.name`| This is the name of the RabbitMQ Cluster CRD, if this is set then the `connectionSecret` must not be set |
-| `rabbitmqClusterReference.connectionSecret.name`| The RabbitMQ Connection Secret name, if this is set then the `clusterReference.name` must not be set |
+| `rabbitmqClusterReference.connectionSecret.name`| The RabbitMQ Connection Secret name. The secret must contain `uri`, `username`, `password`, `port* (default 5672)` (* optional). If this is set then the `clusterReference.name` must not be set |
```
gab-satchi, line 172: The secret must contain `uri`, `username` and `password`. It may optionally contain `port` (default 5672).

gab-satchi (commented):
Just some minor docs readability things. If you can remove the self-signed bits from the description, this is good to go

gab-satchi:
I think we should test the `RabbitMQURL` function. It feels more like a util anyways

gabo1208:
maybe test the rabbitmqurl method in a following pr?

gab-satchi (approved):
/lgtm

gab-satchi (approved):
/lgtm

PR #915, opened 2022-09-02
Guest user available for e2e test in production RabbitMQ Clusters envs

gabo1208:
# Changes

- 🧹 Fixed vhost test case when the available rabbitmq cluster is in producction mode

/kind cleanup


salaboy (approved):
/lgtm

PR #917, opened 2022-09-02
added convert to rabbitmq method to protocol binding + tests

gabo1208:
# Changes

- 🧹 added convert to rabbitmq method to protocol binding + tests

/kind cleanup


gab-satchi (approved):
/lgtm

Just a question around CloudEvents. Do we validate the required CE attributes like spec version, type, source and ID?

gabo1208:
> /lgtm
> 
> Just a question around CloudEvents. Do we validate the required CE attributes like spec version, type, source and ID?

Since here this should be a valid cloud event, we dont do validation. But in the other way around converting from rmq msg to ce we do (in the ingress case the http protocol makes the proper validation when converting to ce)

PR #927, opened 2022-09-08
Adds support for eventing-rabbitmq->RMQ SSL communication with self-signed certs

### LINKED ISSUE #883: Support RMQ Cluster configured with TLS using self-signed certificates
**Problem**
A user configures RMQ cluster with TLS using self-signed certificates and would like to have Knative Eventing securely connect to the RMQ Cluster. 
The two separate pieces would be
- RMQ Cluster -> Knative RMQ Source
- RMQ Cluster <-> Knative RMQ Broker

Which persona is this feature for?
This feature enables Developers to comply with Administrator policies

**Exit Criteria**
- RMQ Cluster is configured with TLS using self signed certificates
- An Eventing RMQ Broker is configured to connect to RMQ over TLS
- Send event to Broker's ingress and the event is delivered to a downstream consumer (with hops between Knative Eventing & RMQ being secured with TLS)

Similar steps for RMQ Source

**Time Estimate (optional):**

This blocked by https://github.com/knative-sandbox/eventing-rabbitmq/issues/566

gab-satchi:
# Changes

- 🎁 Broker -> RabbitMQ TLS communication with self-signed certs
- Requires `RabbitMQCluster.Spec.TLS.CASecretName` to be set to a secret containing the ca.cert
- Requires `caSecretName` as an extra field when using a `connectionSecret` instead of RabbitMQCluster

/kind enhancement

Fixes #883 

**Release Note**

```release-note
- Eventing-rabbitmq can be used with a RabbitMQ server which uses self-signed certs
```

Docs and e2e tests in a follow-up PR

### DIFF
--- a/pkg/rabbit/service.go
+++ b/pkg/rabbit/service.go
@@ -324,8 +324,7 @@ func (r *Rabbit) GetRabbitMQCASecret(ctx context.Context, clusterRef *rabbitv1be
 		if err != nil {
 			return "", err
 		}
-		secretName, _ := s.Data["caSecretName"]
-		return string(secretName), nil
+		return string(s.Data["caSecretName"]), nil
 	}
 
 	rabbitMQCluster, err := r.getClusterFromReference(ctx, clusterRef)
--- a/pkg/rabbit/service.go
+++ b/pkg/rabbit/service.go
@@ -283,33 +283,13 @@ func (r *Rabbit) RabbitMQURL(ctx context.Context, clusterRef *rabbitv1beta1.Rabb
 		return url.Parse(fmt.Sprintf("%s://%s:%s@%s:%s", protocol, username, password, splittedUri[0], port))
 	}
 
-	// TODO: make this better.
-	ref := &duckv1.KReference{
-		Kind:       "RabbitmqCluster",
-		APIVersion: "rabbitmq.com/v1beta1",
-		Name:       clusterRef.Name,
-		Namespace:  clusterRef.Namespace,
-	}
-	gv, err := schema.ParseGroupVersion(ref.APIVersion)
+	rab, err := r.getClusterFromReference(ctx, clusterRef)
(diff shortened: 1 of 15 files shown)


gab-satchi:
@salaboy @xtreme-sameer-vohra Ready for review. 

Most of the ca cert logic is covered by this dial [function](https://github.com/rabbitmq/amqp091-go/blob/5039e19404bbe510a5a0bfde8268055ab5230259/connection.go#L145-L157). When it detects `amqps` which will be part of our URL if TLS is enabled, it will automatically look for system root ca certs to use.

The key change is to load the ca certs into all of the pods that will need to reach out to RabbitMQ. Those are ingress, dispatcher (DLQ and normal Trigger) and adapter. 

The API changes which I will cover in more detail in the docs update are:
- If RabbitmqCluster is used, it already has a caSecretName field in its TLS [Spec](https://github.com/rabbitmq/cluster-operator/blob/97e2dc5d4cb9bfdfe4611ac47273180719f18ac5/api/v1beta1/rabbitmqcluster_types.go#L332). We will use the secret that's supplied there.
- If a Secret is used to connect to RabbitMQ, we require a new field in the data `caSecretName` which will be the name of the secret with the CA certs.

ikvmw:
looks good

salaboy:
This looks good to me too.. I guess to comment on something that @ikvmw  brought in our sync is that we might need to add a bit of documentation for this PR about the secret name for the cert. Besides that, this PR seems to be doing exactly what it is supposed to do.

xtreme-sameer-vohra:
Good call out regarding documentation. @ikvmw Would you be up for tackling the documentation prior to Kn 1.8 release ? It might make sense to create & tackle it as part of a separate issue to make it easier to manage the PR.

ikvmw:
yep, I can do that

PR #947, opened 2022-10-11
Resource request and limits can be set for Broker, Trigger and Source

### LINKED ISSUE #823: Support manually setting CPU requests for dispatcher pods
**Problem**
With release 1.5, CPU requests have been added to the dispatcher pods, but they are not configurable, which is an issue in some cases (see https://github.com/knative-sandbox/eventing-rabbitmq/issues/813#issuecomment-1158525701).

System operator

**Exit Criteria**
CPU request for dispatcher pod can be set to any value, including removing it.

If it's not an option to make this configurable, make sure only one dispatcher pod is needed per deployment consuming events. Currently, when creating a `Trigger` resource, one dispatcher pod is created. There are multiple options to solve this part:

* Supporting the new trigger filters (https://knative.dev/docs/eventing/experimental-features/new-trigger-filters/) which would allow to refactor multiple `Trigger` definitions into one
* Using multiple `Trigger` definitions from one dispatcher pod and consume multiple queues from one dispatcher.
* Probably more options that I don't know about :-)

gab-satchi:
- 🎁 Pod resource requests can be set for Broker, Trigger and RabbitmqSource through annotations

/kind enhancement

Fixes #823 

```release-note
Pod resource requests can be set for Broker, Trigger and RabbitmqSource through annotations
```

### DIFF
--- a/pkg/apis/eventing/v1alpha1/broker_validation.go
+++ b/pkg/apis/eventing/v1alpha1/broker_validation.go
@@ -22,6 +22,7 @@ import (
 
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 
+	"knative.dev/eventing-rabbitmq/pkg/utils"
 	eventingv1 "knative.dev/eventing/pkg/apis/eventing/v1"
 	"knative.dev/pkg/apis"
 	"knative.dev/pkg/apis/duck"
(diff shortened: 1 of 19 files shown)


gabo1208:
The only thing I see that could be improved is that the ingress is receiving the same resource requests/limits than the dispatcher maybe we should adjust that in the future, but for now I like the change

gab-satchi:
Ingress and Dispatcher can be individually adjusted. When the annotation is used on a Broker, it will impact the ingress pod. When it's used on a Trigger it will impact the dispatcher. 

If you meant the DLQ dispatcher then yes. If it's a Broker DLQ, it will be the same as the ingress resources if those were set. If it's a Trigger DLQ, it will use the same resources as the Trigger dispatcher if the annotations are set.

PR #954, opened 2022-10-14
Fix: Specify `securityContext`.

mattmoor:
:bug: Specify a best-practices securityContext for each of the ingress/dispatcher pods we spin up alongside user workloads.

This popped out poking around in the new GKE security posture dashboard

/kind bug

**Release Note**

```release-note
Have ingress/dispatcher specify a more restrictive securityContext
```

**Docs**

```docs
NONE
```

### DIFF
--- a/pkg/reconciler/broker/resources/dispatcher.go
+++ b/pkg/reconciler/broker/resources/dispatcher.go
@@ -32,6 +32,7 @@ import (
 
 	"knative.dev/pkg/apis"
 	"knative.dev/pkg/kmeta"
+	"knative.dev/pkg/ptr"
 	"knative.dev/pkg/system"
 )
 
@@ -176,6 +177,14 @@ func MakeDispatcherDeployment(args *DispatcherArgs) *appsv1.Deployment {
 							Name:          "http-metrics",
 							ContainerPort: 9090,
 						}},
+						SecurityContext: &corev1.SecurityContext{
+							AllowPrivilegeEscalation: ptr.Bool(false),
+							ReadOnlyRootFilesystem:   ptr.Bool(true),
+							RunAsNonRoot:             ptr.Bool(true),
+							Capabilities: &corev1.Capabilities{
+								Drop: []corev1.Capability{"all"},
+							},
+						},
 					}},
 				},
 			},
--- a/pkg/reconciler/broker/resources/ingress.go
+++ b/pkg/reconciler/broker/resources/ingress.go
@@ -30,6 +30,7 @@ import (
 	eventingv1 "knative.dev/eventing/pkg/apis/eventing/v1"
 	reconcilersource "knative.dev/eventing/pkg/reconciler/source"
 	"knative.dev/pkg/kmeta"
+	"knative.dev/pkg/ptr"
 	"knative.dev/pkg/system"
 )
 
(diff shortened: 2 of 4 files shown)


pkg/reconciler/broker/resources/ingress.go
```diff
@@ -30,6 +30,7 @@ import (
 	eventingv1 "knative.dev/eventing/pkg/apis/eventing/v1"
 	reconcilersource "knative.dev/eventing/pkg/reconciler/source"
 	"knative.dev/pkg/kmeta"
+	"knative.dev/pkg/ptr"
```
vaikas, line 33: I wonder if at some point we should deprecate this and use this, looks like they added it and does the same thing?
https://pkg.go.dev/k8s.io/utils/pointer

mattmoor:
Failure doesn't look related to my change, but I can't easily rerun.

mattmoor:
cc @evankanderson 

This is in the same vein as the Serving change FYI.  Not sure who the best folks to ping on Rabbit are these days?

gabo1208:
@mattmoor yo can ping me or @gab-satchi :) thanks for the change we'll review it next week

mattmoor:
@gabo1208 perfect!  No rush, thanks! 🙏

gab-satchi (approved):
/lgtm

thanks for the changes. @gabo1208 note for us to change this for the source-adapter as well

mattmoor:
FWIW, I'd love to get this into the release tomorrow, so we can pick it up and quiet some security alerts 🤞 🙏

pkg/reconciler/broker/resources/ingress.go
imjasonh: I looked into it just now, and k8s pointer doesn't have an equivalent to `ptr.Time`, which [has some usages](https://cs.github.com/?scopeName=All+repos&scope=&q=ptr.Time+knative.dev%2Fpkg%2Fptr) we'll have to remove first. 😕

vaikas: Thanks @imjasonh yeah, wasn't supposed to hold anything up, just popped into my head while looking at this!

mattmoor:
oh nice, that works for actions now :)

pkg/reconciler/broker/resources/dispatcher.go
```diff
@@ -176,6 +177,14 @@ func MakeDispatcherDeployment(args *DispatcherArgs) *appsv1.Deployment {
 							Name:          "http-metrics",
 							ContainerPort: 9090,
 						}},
+						SecurityContext: &corev1.SecurityContext{
+							AllowPrivilegeEscalation: ptr.Bool(false),
+							ReadOnlyRootFilesystem:   ptr.Bool(true),
+							RunAsNonRoot:             ptr.Bool(true),
+							Capabilities: &corev1.Capabilities{
+								Drop: []corev1.Capability{"all"},
```
matzew, line 185: Hi, @mattmoor 

on eventing "core" (and serving), we have replaced `all` with `ALL`, since this is the only acceptable string check, see below:
* https://github.com/kubernetes/kubernetes/blob/6dbec8e25592d47fc8a8269c86d4b5fa838d320b/staging/src/k8s.io/pod-security-admission/policy/check_capabilities_restricted.go#L30 
* https://github.com/kubernetes/kubernetes/blob/6dbec8e25592d47fc8a8269c86d4b5fa838d320b/staging/src/k8s.io/pod-security-admission/policy/check_capabilities_restricted.go#L94

mattmoor: TIL 🤔

mattmoor: https://github.com/knative-sandbox/eventing-rabbitmq/pull/966

PR #992, opened 2022-11-08
Clusterref default namespace

gabo1208:
# Changes

- 🐛 When BrokerConfig Cluster reference was left blank, it didn't defaulted to its parent Broker namespace causing an error with the resources namespace

/kind bug

Fixes #

**Release Note**

```release-note
🐛 When BrokerConfig Cluster reference was left blank, it didn't defaulted to its parent Broker namespace causing an error with the resources namespace
```


pkg/brokerconfig/service.go
```diff
@@ -90,6 +90,9 @@ func (r *BrokerConfigService) GetRabbitMQClusterRef(ctx context.Context, b *even
 		if config.Spec.RabbitmqClusterReference == nil {
 			return nil, fmt.Errorf("rabbitmqBrokerConfig %s/%s spec is empty", b.Spec.Config.Namespace, b.Spec.Config.Name)
 		}
+		if config.Spec.RabbitmqClusterReference.Namespace == "" {
```
gab-satchi, line 93: This won't be possible right since the rabbitmqBrokerConfig has the defaulting that happens above?

gabo1208: We default the brokerconfig in the broker spec, but not the broker config resource per se.

I don't know exactly how to add the defaulting for this one, so any helps is appreciated =)

PR #996, opened 2022-11-10
RabbitmqCluster Connection Secret improvements

gabo1208:
# Changes

- 🧹 Added e2e tests for this feature
- 🎁 Now the "http://" protocol bit can be omitted when using cluster ref connection secret in Broker and Source

/kind enhancement

Fixes #

**Release Note**

```release-note
🎁 Now the "http://" protocol bit can be omitted when using cluster ref connection secret in Broker and Source
```


gab-satchi:
I think we'd need something more descriptive for the release notes instead of 

```
Improved uri info inference when using cluster ref connection secret in Broker and Source
```

The main change seems to be if the http/https protocol was included in the uri, we clean that up. Changes lgtm

PR #1018, opened 2022-12-08
Removes using callbacks in webhooks. Relies on genericCRD interface instead to receive the defaulting and validating calls.

gab-satchi:
# Changes

- :wastebasket: Removes using callbacks in webhooks
- 🗑️ removes extra conversion between unstructured objects to the eventing v1 types
- 🐛 Fixes an issue where an invalid Trigger can be created if its Broker was missing.

```release-note
🐛 Fixes an issue where an invalid Trigger can be created if its Broker was missing.
```

/kind cleanup


pkg/apis/eventing/v1alpha1/broker_validation_test.go
```diff
@@ -302,121 +301,3 @@ func TestValidate(t *testing.T) {
...
 }
-
-func TestValidateFunc(t *testing.T) {
-	tests := []struct {
-		name     string
-		b        *unstructured.Unstructured
-		original *eventingv1.Broker
-		want     *apis.FieldError
-	}{{
-		name: "not my broker class",
```
gab-satchi, line 313: Each of these test cases are covered in the test above. This was testing the wrapper `ValidateBroker` function which is now fully removed.

pkg/apis/eventing/v1alpha1/broker_validation.go
```diff
@@ -119,20 +126,20 @@ func (b *RabbitBroker) Validate(ctx context.Context) *apis.FieldError {
...
 }
 
-func ValidateBroker(ctx context.Context, unstructured *unstructured.Unstructured) error {
-	return validate(ctx, unstructured, &RabbitBroker{})
-}
+func (t *RabbitBroker) SetDefaults(ctx context.Context) {}
 
-func validate(ctx context.Context, unstructured *unstructured.Unstructured, t apis.Validatable) error {
-	if unstructured == nil {
+func (b *RabbitBroker) DeepCopyObject() runtime.Object {
```
gabo1208, line 131: This wasn't autogenerated before?

pkg/apis/eventing/v1alpha1/broker_validation.go
```diff
@@ -20,23 +20,30 @@ import (
...
 	"knative.dev/pkg/apis"
-	"knative.dev/pkg/apis/duck"
 	"knative.dev/pkg/kmp"
+	"knative.dev/pkg/webhook/resourcesemantics"
 )
 
-// +k8s:deepcopy-gen=false
+// stub Broker in order to set up validations and defaults
+// +k8s:controller-gen=false
 type RabbitBroker struct {
```
gabo1208, line 35: So now if the eventing.Broker type changes we have to update this ourselves?

pkg/apis/eventing/v1alpha1/trigger_validation.go
```diff
@@ -21,50 +21,48 @@ import (
...
+	metav1.ObjectMeta `json:"metadata,omitempty"`
 
-// +k8s:deepcopy-gen=false
-type RabbitTrigger struct {
-	eventingv1.Trigger
-	Client versioned.Interface
+	Spec eventingv1.TriggerSpec `json:"spec,omitempty"`
+
+	// +optional
+	Status eventingv1.TriggerStatus `json:"status,omitempty"`
```
gabo1208, line 48: Why is the status optional now?

gab-satchi: This is just a direct copy of the eventing [core api](https://github.com/knative/eventing/blob/main/pkg/apis/eventing/v1/trigger_types.go#L45-L57). I think status has to be optional since it's only set once the controller reconciles the resource.

pkg/apis/eventing/v1alpha1/broker_validation.go
gab-satchi: Yeah both the RabbitBroker and RabbitTrigger will have to track the core API. Spec/Status changes are fine but when the version gets bumped, it'll need changing here and elsewhere in the codebase

pkg/apis/eventing/v1alpha1/broker_validation.go
gab-satchi: No. But good catch. I was able to generate it by adding another comment.

gabo1208:
Let me know when this is ready @gab-satchi

gab-satchi:
It's ready @gabo1208

gabo1208 (approved):

PR #1073, opened 2023-02-23
don't stop polling if there is a not found error

gabo1208:
# Changes

- :broom: Don't stop polling for the secret if there is a not found error in the e2es

/kind cleanup


test/e2e/rabbitmqcluster.go
```diff
@@ -118,7 +118,11 @@ func patchConnectionSecret(ctx context.Context, namespace string, secretName str
 	err = wait.PollImmediate(interval, timeout, func() (bool, error) {
 		secret, err = kubeClient.Get(ctx).CoreV1().Secrets(namespace).Get(ctx, secretName, metav1.GetOptions{})
 		if err != nil {
-			return false, err
+			if apierrors.IsNotFound(err) {
```
kvmware, line 121: if it is another error then it would not return false which could cause issues because it is hiding the error

PR #1077, opened 2023-02-23
adding stop clause to poll inmediate

gabo1208:
# Changes

- 🧹 Now if there is any error other than not found in the secrets polling, it stops and report it

/kind cleanup


gab-satchi (approved):
Thanks. 

/lgtm

PR #1138, opened 2023-06-02
removing non used env var for the receive adapter and old source docs…

gabo1208:
improving coverage
adding little nits fixes


gabo1208:
cc @kvmware @dprotaso a bit of help with this one

PR #1163, opened 2023-07-10
Delivery failure

### LINKED ISSUE #1154: RabbitMQ Trigger knativeerrordest extension
**Problem**
When a trigger fails and deadLetterSink is setup, It is not possible to see what Trigger (Service) failed on a specific event, because Extension in the Event is not set. 

Event consumer (developer)
System Integrator

**Exit Criteria**
When the failed event is recived by the deadLetterSink it must contain knativeerrordest in the Event Extension.

**Time Estimate (optional):**

Add any other context about the feature request here.

thomasjohansen:
# Changes

- :gift: Add delivery failure (like for channels: https://knative.dev/docs/eventing/event-delivery/#configuring-channel-event-delivery)

/kind enhancement

Fixes #1154

**Release Note**

```release-note
Added error information to failed messages
```

### DIFF
--- a/pkg/dispatcher/dispatcher.go
+++ b/pkg/dispatcher/dispatcher.go
@@ -49,6 +49,7 @@ type Dispatcher struct {
 	BrokerIngressURL  string
 	SubscriberURL     string
 	SubscriberCACerts string
+	DeadLetterSinkURL string
 	MaxRetries        int
 	BackoffDelay      time.Duration
 	Timeout           time.Duration
@@ -152,6 +153,8 @@ func getStatus(ctx context.Context, result protocol.Result) (int, bool) {
 
 func (d *Dispatcher) dispatch(ctx context.Context, msg amqp.Delivery, ceClient cloudevents.Client) {
 	start := time.Now()
+	subscriberURL := d.SubscriberURL
+
 	msgBinding := rabbit.NewMessageFromDelivery(ComponentName, "", "", &msg)
 	event, err := binding.ToEvent(cloudevents.WithEncodingBinary(ctx), msgBinding)
 	if err != nil {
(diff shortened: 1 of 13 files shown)


pkg/dispatcher/dispatcher.go
```diff
@@ -180,25 +190,33 @@ func (d *Dispatcher) dispatch(ctx context.Context, msg amqp.Delivery, ceClient c
 	statusCode, isSuccess := getStatus(ctx, result)
 	if statusCode != -1 {
 		args := &dispatcher.ReportArgs{EventType: event.Type()}
-		if err := d.Reporter.ReportEventCount(args, statusCode); err != nil {
+		if err = d.Reporter.ReportEventCount(args, statusCode); err != nil {
 			logging.FromContext(ctx).Errorf("Something happened: %v", err)
+
+			msg.Headers["knativeerrordest"] = d.SubscriberURL
```
gabo1208, line 196: And you can remove it from here, this is just knative related so if it fails it shouldn't affect the user

pkg/dispatcher/dispatcher.go
```diff
@@ -150,15 +151,22 @@ func getStatus(ctx context.Context, result protocol.Result) (int, bool) {
...
+
 	msgBinding := rabbit.NewMessageFromDelivery(ComponentName, "", "", &msg)
 	event, err := binding.ToEvent(cloudevents.WithEncodingBinary(ctx), msgBinding)
 	if err != nil {
-		logging.FromContext(ctx).Warn("failed creating event from delivery, err (NACK-ing and not re-queueing): ", err)
-		err = msg.Nack(false, false)
-		if err != nil {
-			logging.FromContext(ctx).Warn("failed to NACK event: ", err)
+		logging.FromContext(ctx).Warn("failed parsing event, setting knativeerrordest header and NACK-ing and re-queueing: ", err)
+
```
gabo1208, line 167: Here we are missing also the errordest set and we can send an empty event in this case

pkg/dispatcher/dispatcher.go
```diff
@@ -150,15 +151,22 @@ func getStatus(ctx context.Context, result protocol.Result) (int, bool) {
 	return -1, false
 }
 
-func (d *Dispatcher) dispatch(ctx context.Context, msg amqp.Delivery, ceClient cloudevents.Client) {
+func (d *Dispatcher) dispatch(ctx context.Context, msg amqp.Delivery, ceClient cloudevents.Client, channel rabbit.RabbitMQChannelInterface) {
 	start := time.Now()
+	if _, ok := msg.Headers["knativeerrordest"]; ok {
+		if err := msg.Ack(false); err != nil {
```
gabo1208, line 157: Here we want to nack, since we want the message to go to the DLQ

pkg/reconciler/trigger/resources/dispatcher.go
```diff
@@ -198,6 +198,13 @@ func MakeDispatcherDeployment(args *DispatcherArgs) *appsv1.Deployment {
 				Value: *args.Subscriber.CACerts,
 			})
 	}
+	if args.DLX {
+		dispatcher.Env = append(dispatcher.Env,
+			corev1.EnvVar{
+				Name:  "DLX",
+				Value: "true",
```
gabo1208, line 205: Since this is the Trigger's Dispatcher his should have the value of args.DLX

This one you can put it on the pkg/reconciler/brokker/resources/dispatcher.go that is always going to be true

pkg/dispatcher/dispatcher.go
```diff
@@ -217,6 +223,128 @@ func (d *Dispatcher) dispatch(ctx context.Context, msg amqp.Delivery, ceClient c
 	}
 }
 
+func (d *Dispatcher) dispatchDLQ(ctx context.Context, msg amqp.Delivery, ceClient cloudevents.Client, channel rabbit.RabbitMQChannelInterface) {
+	start := time.Now()
+	if _, ok := msg.Headers["knativeerrordest"]; ok {
```
gabo1208, line 228: In the case of the dispatch DLQ method, remember that is the code we had before your changes, so this function should be the other code

gabo1208 (changes requested):

pkg/reconciler/trigger/resources/dispatcher.go
thomasjohansen: Env variable will only be set if DLX is true. Defaults to false

pkg/dispatcher/dispatcher.go
```diff
@@ -104,9 +104,9 @@ func (d *Dispatcher) ConsumeFromQueue(ctx context.Context, conn rabbit.RabbitMQC
 			defer wg.Done()
 			for msg := range workerQueue {
 				if d.DLX {
-					d.dispatchDLQ(ctx, msg, ceClient)
```
gabo1208: this is required?

gabo1208 (approved):

PR #1173, opened 2023-07-18
Fixed wrong Source's webhook service yaml definition

### LINKED ISSUE #1128: rabbitmq-webhook-certs specified twice in release manifest
**Describe the bug**

The released rabbitmq-source.yaml has the rabbitmq-webhook-certs Secret specified twice, which flux/kustomize chokes on.

 **Expected behavior**

The rabbitmq-webhook-certs is only  specified once.

 **Knative release version**

1.10.0

gabo1208:
… causing some instalations to fail, now it has a propper one

# Changes

- 🐛 Fixed wrong Source's webhook service yaml definition

/kind bug

Fixes #1128

### DIFF
--- a/config/source/400-webhook-service.yaml
+++ b/config/source/400-webhook-service.yaml
@@ -13,10 +13,17 @@
 # limitations under the License.
 
 apiVersion: v1
-kind: Secret
+kind: Service
 metadata:
-  name: rabbitmq-webhook-certs
-  namespace: knative-sources
   labels:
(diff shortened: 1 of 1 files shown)


ChunyiLyu (approved):
LGTM. Thanks for the fix! For context, rmq webhook certs secret are already defined in https://github.com/knative-sandbox/eventing-rabbitmq/blob/main/config/source/500-webhook-secret.yaml. This is likely due to a copy paste error.

PR #1239, opened 2023-09-29
removing parse url from the get rabbitmqUrl function

### LINKED ISSUE #1216: Not able to connect to RabbitMQ with a virtual host configured.
**Describe the bug**
I am not able to connect to an external rabbitMQ cluster which is accessible through my C# code. In the external connection string configuration, I have not found a place to provide rabbitMQ virtual host. I think it is mandatory for me to provide that to connect to the external rabbitMQ instance as I have access only to that virtual host. 

 **Expected behavior**
I should see a connection established to rabbitMQ

 **To Reproduce**
1. Create a virtual host in a rabbitMQ cluster
2. Create a user and give that user access to that virtual user
3. Provide connection details in rabbitMQ secret file 
4. Use RabbitMQSource to connect to rabbitMQ by providing the previously created secret.

 **Knative release version**
latest

gabo1208:
# Changes

- 🎁 Now Brokers and Sources accept broader rabbitmq url usernames and passwords characters to connect to RabbitMQ instances

/kind bug

Fixes #1216

**Release Note**

```release-note
Now Brokers and Sources accept broader rabbitmq url usernames and passwords characters to connect to RabbitMQ instances
```

### DIFF
--- a/pkg/rabbit/exchange.go
+++ b/pkg/rabbit/exchange.go
@@ -17,8 +17,6 @@ limitations under the License.
 package rabbit
 
 import (
-	"net/url"
-
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"knative.dev/eventing-rabbitmq/pkg/apis/sources/v1alpha1"
 	"knative.dev/pkg/kmeta"
@@ -33,7 +31,7 @@ type ExchangeArgs struct {
 	Namespace                string
 	RabbitMQVhost            string
 	RabbitmqClusterReference *rabbitv1beta1.RabbitmqClusterReference
-	RabbitMQURL              *url.URL
+	RabbitMQURL              string
 	Broker                   *eventingv1.Broker
 	Trigger                  *eventingv1.Trigger
 	Source                   *v1alpha1.RabbitmqSource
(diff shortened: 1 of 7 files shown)


gabo1208:
@kvmware  @dprotaso here I need to ignore the test coverage since it is not affecting anything reported by codecov

dolfolife (commented):
I wonder if doing `url.QueryEscape` on the username might be better than us not constructing (validating a URL)
For example https://go.dev/play/p/ol-loHQ3pRn

ChunyiLyu (approved):
Just to check my understanding: the issue was user had a back slash in their credentials which is parsed as a special case by `url.Parse` and you’ve fixed it by removing `url.Parse`? 

If you want to keep validating the URl as before, you could consider 1) adding `username` and `password` as separate fields to `ExchangeArgs` and keep `RabbitMQURL` as it is 2) construct the URL with `url.URL` first before converting it to string. I'm not sure if the URL is validated anywhere else. If it isn't, keeping the validation would be a safer option. You can take a look of messaging topology operator as a reference: https://github.com/rabbitmq/messaging-topology-operator/blob/main/rabbitmqclient/cluster_reference.go

It's been a while since I've looked at eventing. If the team is happy with the change please feel free to disregard my comment and go ahead 😀

PR #1372, opened 2024-04-05
Drop performance tests

dprotaso:
mako was a Google tool but it's no longer hosted and maintained so figure we can remove it for now 

Part of https://github.com/knative/pkg/issues/2820


evankanderson:
Sorry, just catching this now, but I don't think this was using `mako` at all?  It looks like it was using `gnuplot` and some manual steps.

It would be nice to bring these back in a cross-Eventing way in some form, rather than having them buried in the `rabbitmq` implementation, but I wanted to make sure that I understood whether these were using Mako at all or not.

gabo1208:
We were using [mako](https://github.com/knative-extensions/eventing-rabbitmq/pull/1372/commits/a6e0f84d004a53607e80ac5f1a10a726a01b3bbc#diff-bc937250a4f271533a2098b30e65327f080a017da7b32c9c0a985020144dee5dL100) as a sidecar to knative perf test images, so maybe we didn't need mako at all, or just for some aggregation.

Should be not that hard to reuse the same images and aggregate, and just automating something like [this results script](https://github.com/knative-extensions/eventing-rabbitmq/pull/1372/commits/a6e0f84d004a53607e80ac5f1a10a726a01b3bbc#diff-bea0df81e6ece5471ea1adbd1dc1f112d3aa31b7a448288d96a270a16620d909L171)

PR #1381, opened 2024-04-16
Drop mako-stub

Zerpet:
# Changes

- :wastebasket: Remove mako-stub

/kind removal

Related https://github.com/knative/pkg/issues/2820


ikavgo:
@Zerpet there is an CLA thingy going on, could you please have a look

PR #1422, opened 2024-06-20
drop makostub dependency

dprotaso:
Missed the line in https://github.com/knative-extensions/eventing-rabbitmq/pull/1372


pierDipi (approved):
/lgtm
/approve

PR #1440, opened 2024-07-23
Fix the dependency update for Knative 1.15 release

Leo6Leo:
# Changes

- Fix the permissions for the broker controller by adding the eventpolicies roles so that it can start


dsimansk:
@Leo6Leo the error you posted in your last comment is Golang version related. IMO it's hardcoded in GH action to be 1.21 or similar.

PR #1589, opened 2025-07-23
go mod vendor

gauron99:
run `go mod vendor`


gauron99:
this was just an initial test if a simple re-do of the old missing lib was enough. Seems like it will require more comprehensive review and redo

PR #1590, opened 2025-07-23
Migrate to OTel

Cali0707:
# Changes

- Migrate eventing-rabbitmq to OTel
- Update dependencies to get new required deps

**Release Note**

```release-note
eventing-rabbitmq now uses OpenTelemetry for metrics and traces
```


pkg/dispatcher/dispatcher.go
```diff
@@ -196,12 +229,6 @@ func (d *Dispatcher) dispatch(ctx context.Context, msg amqp.Delivery, ceClient c
 
 	response, result := ceClient.Request(ctx, *event)
 	statusCode, isSuccess := getStatus(ctx, result)
-	if statusCode != -1 {
-		args := &dispatcher.ReportArgs{EventType: event.Type()}
-		if err = d.Reporter.ReportEventCount(args, statusCode); err != nil {
```
evankanderson, line 201: Did we stop reporting counts of errors?  That feels like a useful metric to keep.

It's also possible the defer is getting the http error somehow -- if so, that's probably a detail worth commenting, since earlier code is depending on the actions of later code, rather than the usual.

pkg/utils/envconfig.go
```diff
@@ -67,33 +59,71 @@ func (e *EnvConfig) GetLogger() *zap.SugaredLogger {
...
+	otel.SetTextMapPropagator(tracing.DefaultTextMapPropagator())
+	otel.SetTracerProvider(tracerProvider)
+
+	e.meterProvider = meterProvider
+	e.traceProvider = tracerProvider
+
 	return nil
 }
+
+func (e *EnvConfig) ShutdownObservability(ctx context.Context) {
```
evankanderson, line 120: I like having this encapsulated in a function, rather than raw in the main file.  Thanks!

evankanderson (commented):
One question, otherwise looks good. 

/approve

ikavgo:
should `latencyBounds` be configurable?

evankanderson (commented):
Yes, I got here late. 

I think `latencyBounds` are actually `latencyBuckets`, and it's reasonable for them to be fixed at compile time -- when multiple servers report using different bucket sizes, it's very hard to produce useful histograms later.

ikavgo:
If they are hardcoded why we sure multiple servers have them the same?

Cali0707:
> If they are hardcoded why we're sure multiple servers have them the same?

In OTel, it is normally done by convention -  we have been following the semantic convention for http request duration for the event dispatch duration metric (see [here](https://opentelemetry.io/docs/specs/semconv/http/http-metrics/#metric-httpserverrequestduration))

PR #1760, opened 2026-06-22
Fix RabbitMQ  connection close vs HTTP drain race during ingress shutdown (#1669) 

### LINKED ISSUE #1669: Broker ingress closes connection to rabbitmq to early
**Describe the bug**
We have seen that the broker ingress will stop the rabbitmq connections but is still accepting HTTP requests. 
This ends in a lot of errors during this shutdown phase. 

 **Expected behavior**
The ingress stops first accepting HTTP requests (readiness probe?) and stops the rabbitmq connections at the very end of the shutdown

 **To Reproduce**
Send cloudevents using http to the ingress and then trigger a restart of the ingress

 **Knative release version**
1.20.x

 **Additional context**

Image

deadtrickster:
# Changes

:bug #1669 fixes RabbitMQ connection closure vs Drain race.

/kind bug

Fixes #1669

**Release Note**

```release-note
Fix RabbitMQ  connection close vs HTTP drain race during ingress shutdown (#1669) 
```

@dprotaso it's been some time since I contributed here. Could you please have a look if I'm doing the right thing here?

Two things worth noting - had to update kind version and the e2e test - it looks like the standard mechanics sends message one by one and rather infrequently (I suspect 1msg/sec) so it is hard to trigger the situation that @eloo-abi had.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -10,3 +10,6 @@ bin/
 .envrc
 .env
 .config
+
+# transient test output
+*.log
--- a/Makefile
+++ b/Makefile
@@ -136,7 +136,7 @@ releases-envsubst:
 
 KUBECTL_RELEASES := https://github.com/kubernetes/kubernetes/tags
 # Keep this in sync with KIND_K8S_VERSION
-KUBECTL_VERSION := v1.34.3
+KUBECTL_VERSION := v1.34.4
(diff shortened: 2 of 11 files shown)


deadtrickster:
I also noticed controller-gen is rather old. shall I update here or separately?

